PR #241, opened 2026-04-02
Fix: helpful error when configured server is unreachable

### LINKED ISSUE #240: Improve error message when configured server is unreachable
## Problem

When a user has a server configured but it's unreachable, CLI commands like `nebi publish` and `nebi registry add` fail with a raw connection refused error:

```
Error: creating registry: Post "http://localhost:8460/api/v1/admin/registries": dial tcp [::1]:8460: connect: connection refused
```

This is confusing because:
1. Users don't know if they need to start a server or if something else is wrong
2. The `--local` flag workaround isn't discoverable from the error
3. Users who previously logged in but no longer need the server don't know `nebi logout` would fix it

## Expected behavior

When a server is configured but unreachable, the CLI should provide a helpful error message, e.g.:

```
Error: server at http://localhost:8460 is unreachable.

Hints:
  - If the server should be running, check that it's started
  - To work in local mode instead, use the --local flag
  - To remove the configured server, run: nebi logout
```

## Context

From discussion with @khuyentran1401 after testing PR #222. The CLI correctly defaults to local mode when *no* server is configured, but when a server *is* configured and unreachable, the error is unhelpful. See #239 for the original report.

Adam-D-Lewis:
## Summary
- When a configured server is unreachable, CLI commands now show a clear error with actionable hints instead of a raw `connection refused` / DNS error
- Adds `ConnectionError` type and `wrapConnectionError` helper in `cliclient` that detects network-level failures (`net.OpError`, `net.DNSError`, timeout) and wraps them
- Applied at both `httpClient.Do()` call sites (`request()` and `GetText()`) so all API calls benefit

**Before:**
```
Error: creating registry: Post "http://localhost:8460/api/v1/admin/registries": dial tcp [::1]:8460: connect: connection refused
```

**After:**
```
Error: creating registry: server at http://localhost:8460 is unreachable: ...

Hint: If the server should be running, check that it is started.
  To run this command in local mode instead, use the --local flag.
  To remove the configured server, run: nebi logout
```

Closes #240

## Test plan
- [x] Unit tests for `wrapConnectionError` covering connection refused, DNS errors, non-network errors, and API errors
- [x] `go test ./internal/cliclient/...` passes
- [x] `go test ./internal/store/... ./cmd/nebi/...` passes
- [x] `go vet ./...` clean

### DIFF
--- a/internal/cliclient/client.go
+++ b/internal/cliclient/client.go
@@ -191,7 +191,7 @@ type ConnectionError struct {
 
 func (e *ConnectionError) Error() string {
 	return fmt.Sprintf(
-		"server at %s is unreachable: %v\n\nHint: If the server should be running, check that it is started.\n  To work in local mode instead, use the --local flag.\n  To remove the configured server, run: nebi logout",
+		"server at %s is unreachable: %v\n\nHint: If the server should be running, check that it is started.\n  To run this command in local mode instead, use the --local flag.\n  To remove the configured server, run: nebi logout",
 		e.ServerURL, e.Err,
 	)
 }
--- a/internal/cliclient/client.go
+++ b/internal/cliclient/client.go
@@ -5,9 +5,12 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
+	"net/url"
 	"strings"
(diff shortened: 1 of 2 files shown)


Adam-D-Lewis:
I'm gonna close this because it's super out of date now

PR #278, opened 2026-04-14
Fix redundant package management UX (#218)

tylerpotts:
Closes #218

## Summary

- **Configuration tab is now the single entry point for package management.** Saving the pixi.toml triggers a `pixi install` job ("Save & Install"), which solves the environment, syncs packages to the database, and creates a new version snapshot.
- **Packages tab is now read-only** — displays installed packages without install/delete buttons. Subtitle directs users to the Configuration tab.
- **Edit workspace screen removed** from the workspace list page, along with the per-row Edit button. Configuration editing lives on the workspace detail's Configuration tab.
- **Recreate workspace feature removed** — no longer needed since Save & Install updates the environment in place.
- **TOML Mode is the default** editor mode (moved to the left of UI Mode).
- **Tab order reordered** to: Overview | Configuration | Versions | Packages | Jobs | ...

### Backend
- `SolveEnvironment` added to `Executor` interface and `LocalExecutor` (runs `pixi install -v`)
- `JobTypeUpdate` handled in the worker (solve, sync packages, create version snapshot)
- `POST /workspaces/:id/solve` endpoint with RBAC write permission and audit logging

### Frontend
- `solveWorkspace()` API client method
- Packages tab: removed install form, delete buttons, and Actions column
- Workspaces list: removed edit card, edit button, and all related state/handlers
- Configuration tab: simplified to Save & Install flow only

## Test plan
- [x] Edit pixi.toml on Configuration tab, click Save & Install — verify job runs, packages sync, version created
- [x] Verify Packages tab is read-only (no install/delete buttons)
- [x] Verify no Edit button on workspace list page
- [x] Verify TOML Mode is the default when clicking Edit on Configuration tab
- [x] Verify tab order: Overview | Configuration | Versions | Packages | Jobs | ...

PR #282, opened 2026-04-15
Decouple editor modes with server reload and discard prompt instead of live TOML sync

marcelovilla:
## Reference Issues or PRs
Closes #243 

## Description

As mentioned in #243, the original TOML was not shown when editing an existing workspace, but rather a predefined template with just the dependencies. This PR fixes that by loading the original TOML in the editor from the server.

However, that introduces another issue: both editor modes (i.e. UI and TOML) are not in sync anymore, because the TOML is not being built from a list of packages but it's being loaded from the server. While it was discussed that we would do client-side TOML manipulation to allow syncing both modes, I decided against it after exploring that approach in more detail. It seems fragile and prone to subtle divergence between Pixi's TOML handling and any handrolled client-side approach, especially around comments, inline annotations, and formatting. Pixi preserves all of these via Rust's `toml_edit` crate, which doesn't seem to have an equivalent in the JS/TS ecosystem.

Rather than attempting to keep both modes in sync, this PR decouples them. Each mode loads its state independently (UI mode parses the dependency list, TOML mode loads the raw file from the server). In practice, users choosing one mode are unlikely to need the other mid-session. For the cases where they do switch, a confirmation dialog now warns about unsaved changes that will be lost, making the tradeoff explicit rather than silently discarding work.

Here's a couple demos of the proposed workflow.

**Creating a new workspace:**

https://github.com/user-attachments/assets/ad81607d-1f02-4165-82a7-8374f7ca42b4

**Updating an existing workspace**:

https://github.com/user-attachments/assets/4f3210cd-3cc8-4ff5-92b4-8be36ce63435

### DIFF
--- a/frontend/src/components/workspace/PixiTomlEditor.tsx
+++ b/frontend/src/components/workspace/PixiTomlEditor.tsx
@@ -1,8 +1,9 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 import { Button } from '@/components/ui/button';
 import { Input } from '@/components/ui/input';
 import { Textarea } from '@/components/ui/textarea';
-import { Plus, Trash2, FileCode } from 'lucide-react';
+import { ConfirmDialog } from '@/components/ui/confirm-dialog';
+import { Plus, Trash2, FileCode, Loader2 } from 'lucide-react';
 
 interface Package {
   name: string;
@@ -12,7 +13,8 @@ interface Package {
 interface PixiTomlEditorProps {
   tomlValue: string;
   onTomlChange: (toml: string) => void;
-  workspaceName: string;
+  workspaceName?: string;
+  onReloadToml?: () => Promise<string>;
 }
 
 const buildPixiToml = (packages: Package[], wsName: string): string => {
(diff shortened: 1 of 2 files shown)


kcpevey:
I dont think I like the two extra buttons on existing workspaces. I'd rather change the "create workspace" to "save and create workspace" or "save and create". I think Vini might have actually been using that in a separate effort?

marcelovilla:
@kcpevey do you mean the cancel and the save button on the existing workspaces? For what it's worth, the scope of this PR is outside of those buttons as they were already there and they don't have anything to do with the TOML being lost.

kcpevey:
Reposting this here so it's not lost, and I've thought about it more:
I actually didn't realize the cancel and save buttons on the existing workspace were already existing. 
image


For what it's worth, I think that the UX changes in this PR (pop up dialog warning folks) needs to go along with button change. I realize now that my suggestion of just changing the wording on the "create workspace" button isn't feasible because those two buttons already exist and it would require us to remove them. 

All that said, it doesn't need to be included here, but it probably deserves a UX issue - and I know Tyler is working on other related UX changes so, I'll just step back and see where all that lands.

marcelovilla:
@kcpevey I agree there are UX changes to be made. https://github.com/nebari-dev/nebi/pull/278 already incorporates your suggestions regarding the button labels. @viniciusdc also suggested yesterday that the cancel and workspace buttons should go to the lower right corner to avoid cluttering the UI and I agree. I think some of these also need to be discussed with @smeragoel

aktech (approved):

Issue #284, opened 2026-04-15
OCI workspace bundle format spec


aktech:
Spec for self-contained OCI bundles. Replaces ad-hoc two-layer format. Fully backward compatible — old artifacts are valid bundles with zero assets.

Relates to #268.

## Format

Standard OCI Image Manifest. Config blob unchanged (`application/vnd.pixi.config.v1+toml`, `{}`).

Layers:

<img width="1154" height="518" alt="Image" src="https://github.com/user-attachments/assets/4dab6975-4d85-40be-81ab-ff7fc9335cd9" />

| # | Media type | Title annotation | Count |
|---|---|---|---|
| 0 | `application/vnd.pixi.toml.v1+toml` | `pixi.toml` | 1 (required) |
| 1 | `application/vnd.pixi.lock.v1+yaml` | `pixi.lock` | 1 (required) |
| 2..N | `application/vnd.nebi.asset.v1` | relative path | 0+ |

- No roles. No per-file metadata. Assets are opaque bytes at a path.
- Old 2-layer artifacts = valid bundles with zero asset layers. No version bump.
- Nebi-artifact discriminator stays the config media type (used by `IsNebiRepository`).

### `nebi publish --local`

```
nebi publish --local <workspace-dir> --registry <name> [--tag <tag>] [--concurrency N]
```

- `--local` runs client-side bundling. Without `--local`, existing server-driven publish is unchanged and emits zero-asset bundles.
- `--concurrency` bounds parallel blob pushes. Default 8.

Bundle contents = workspace dir, filtered:
1. Hardcoded drop: `.git/`, `.pixi/`.
2. If `[tool.nebi.bundle].include` is set, keep only files matching those globs. If absent, all files are candidates.
3. Drop anything matched by `.gitignore` (if present).
4. Drop anything matched by `[tool.nebi.bundle].exclude` in `pixi.toml`.
5. Force include: `pixi.toml`, `pixi.lock` (override any include/exclude rule).
6. Symlinks, device files, fifos: skipped silently. Regular files only.

`include` and `exclude` work together: `include` whitelists candidates, `exclude` further filters them down. When only `exclude` is set, everything is a candidate minus the excluded paths (current default).

Config:
```toml
[tool.nebi.bundle]
include = ["src/**", "assets/**", "README.md"]
exclude = ["*.log", "secrets/**", "notes.md"]
```
Unknown keys under `[tool.nebi.bundle]` silently ignored (warning in verbose mode).

### `nebi import`

[...]

PR #295, opened 2026-04-21
feat(oci): workspace bundle format with asset layers

aktech:
Implements #284.

- Bundles now carry pixi.toml, pixi.lock, plus arbitrary asset layers (`application/vnd.nebi.asset.v1`)
- Zero-asset bundles remain byte-compatible with legacy two-layer artifacts
- `nebi publish --local` walks the workspace (`.gitignore` + `[tool.nebi.bundle].include`/`exclude` + hardcoded `.git`/`.pixi` drops); `nebi import` restores assets to the output directory
- Parallel blob transfer (`--concurrency`, default 8)
- Shared path-safety validator runs on both sides
- Unit, integration, and CLI e2e tests (in-memory go-containerregistry)


internal/oci/publisher.go
```diff
@@ -22,144 +23,401 @@ const (
...
 	}
-	if err := repo.Push(ctx, configDesc, configReader); err != nil {
-		return "", fmt.Errorf("failed to push config to registry: %w", err)
+	if err := pushBlobsParallel(ctx, fs, remoteRepo, blobs, cfg.concurrency, cfg.progress); err != nil {
+		return PublishResult{}, err
 	}
 
-	// Push pixi.toml blob
-	tomlReader, err := fs.Fetch(ctx, tomlDesc)
+	manifestReader, err := fs.Fetch(ctx, manifestDesc)
```
nenb, line 330: `defer manifestReader.Close()` (?)

internal/oci/publisher.go
```diff
@@ -22,144 +23,401 @@ const (
...
+// concurrently.
+func WithProgress(fn func(label string, pushed, total int)) PublishOption {
+	return func(c *publishConfig) { c.progress = fn }
+}
+
+// WithAssets bypasses the workspace walker and publishes the supplied
+// list verbatim. Any pixi.toml / pixi.lock entries in the slice are
+// silently dropped — those always ride as typed layers 0 and 1. Unsafe
+// paths still abort pre-network.
+func WithAssets(assets []Asset) PublishOption {
```
nenb: `Asset.AbsPath` is not validated like `Asset.RelPath` is with `validateAssetPaths(paths)`. Think this could be risky if we ever allowed user-controlled input to flow into `WithAssets` in the  future? Could we change the name to `withAssets` and so unexport it and reduce the risk?

internal/oci/publisher.go
```diff
@@ -22,144 +23,401 @@ const (
...
+		desc, err := fs.Add(ctx, asset.RelPath, MediaTypeNebiAsset, asset.AbsPath)
+		if err != nil {
+			return PublishResult{}, fmt.Errorf("cannot read %s: %w", asset.RelPath, err)
+		}
+		desc.Annotations = map[string]string{ocispec.AnnotationTitle: asset.RelPath}
+		layers = append(layers, desc)
+		assetDescs = append(assetDescs, desc)
 	}
 
-	// Create config descriptor (empty JSON)
```
nenb, line 75: GPT comment: `configData` is JSON but `MediaTypePixiConfig` is TOML. 

Not sure how much backward compatability we have to support here, but probably this needs to be sorted.

cmd/nebi/publish.go
nenb: Should we be creating the tag based on more than the `pixiToml`? What happens if I change an asset but don't change the Pixi TOML?

internal/oci/browser.go
```diff
@@ -172,68 +227,220 @@ func ListTags(ctx context.Context, repoRef string, opts BrowseOptions) ([]TagInf
...
+		case MediaTypePixiLock:
+			if haveLock {
+				return out, fmt.Errorf("invalid bundle: duplicate core layer")
+			}
+			out.pixiLock = layer
+			haveLock = true
+		case MediaTypeNebiAsset:
+			out.assets = append(out.assets, layer)
+		default:
+			// Unknown media type — ignore for forward compat.
```
nenb: GPT raised an issue here. We are not validating unknown layers, but we then copy the unknown layers later with `oras.Copy(ctx, repo, tag, fs, tag, copyOpts)` in `ExtractBundle`.

I don't think we want to allow this behaviour - correct? (ie having these unknown layers silently being downloaded, and not being validated)

internal/oci/browser.go
```diff
@@ -172,68 +227,220 @@ func ListTags(ctx context.Context, repoRef string, opts BrowseOptions) ([]TagInf
...
+// path. destDir is created if missing; caller is responsible for the
+// "empty destination" policy (see cmd/nebi/import.go).
+//
+// Implementation delegates blob transfer to oras.Copy over a file.Store
+// rooted at destDir. That gives us streamed I/O, parallel fetches
+// configured via opts.Concurrency, and file.Store's
+// AllowPathTraversalOnWrite=false default as defense-in-depth over the
+// pre-validated titles.
+func ExtractBundle(ctx context.Context, repoRef, tag, destDir string, opts PullOptions) (*PullResult, error) {
+	repo, cm, err := resolveBundleManifest(ctx, repoRef, tag, opts)
```
nenb, line 425: GPT was flagging a TOCTOU bug here - by the time we call `oras.Copy(ctx, repo, tag, fs, tag, copyOpts)` later, the tag could move or someone could push some unknown layers, and then we might extract content different from what we validated

I think it's pretty unlikely, but has a straightforward fix (resolve once, keep the resolved descriptor, and copy by immutable digest/reference from that descriptor instead of by tag) so think it's better to fix now.

internal/oci/browser.go
```diff
@@ -363,18 +570,12 @@ func ChangeRepositoryVisibility(ctx context.Context, host, repoPath, apiToken st
...
 	}
 	defer reader.Close()
-
-	data, err := io.ReadAll(reader)
-	if err != nil {
-		return "", err
-	}
-
-	return strings.TrimSpace(string(data)), nil
+	return io.ReadAll(reader)
```
nenb, line 633: Nit: All uses of `io.ReadlAll` are unbounded reads in this file. In practice I don't think it's an issue, but maybe we want to enforce max sizes.

internal/oci/browser.go
nenb, line 164: GPT flagged that Quay API is paginated and that this fetches only one page and so potentially could miss repositories here unless we loop until no next page.

internal/oci/browser.go
```diff
@@ -172,68 +227,220 @@ func ListTags(ctx context.Context, repoRef string, opts BrowseOptions) ([]TagInf
...
+		case MediaTypeNebiAsset:
+			out.assets = append(out.assets, layer)
+		default:
+			// Unknown media type — ignore for forward compat.
+		}
+	}
+	if !haveToml || !haveLock {
+		return out, fmt.Errorf("invalid bundle: missing pixi.{toml,lock}")
+	}
+	// Validate every asset title before fetch. This also rejects dupes
```
nenb, line 314: We don't validate core layer titles ie `pixi.toml` and `pixi.lock` - should we? Also do we need to make sure that someone doesn't also try and bundle `pixi.toml` here ie a name collision?

internal/oci/walker.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+		relFS, err := filepath.Rel(absRoot, path)
+		if err != nil {
+			return err
+		}
+		rel := filepath.ToSlash(relFS)
+
+		// Hardcoded drops at any depth (top-level only per spec wording,
+		// but treating .git/.pixi as always-drop at any depth is safer).
+		if d.IsDir() {
```
nenb: GPT pointed out that `.git` can be a file for worktrees and submodules.

I don't know how we want to handle this - is a comment and TODO okay for now, or do we think that this is a security footgun big enough to try and prevent now?

internal/oci/walker.go
```diff
@@ -0,0 +1,264 @@
...
+		return nil, err
+	}
+
+	// Ensure force-included files are present even if filters elided them.
+	for name := range forceIncludeFiles {
+		if _, ok := seen[name]; ok {
+			continue
+		}
+		abs := filepath.Join(absRoot, name)
+		info, err := os.Stat(abs)
```
nenb: GPT points out that `os.Stat` follows symlinks and so we should use `os.Lstat` to reject symlinks here (and potentially also validate that resolved paths stay under `absRoot`)

internal/oci/walker.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+	}
+	return ignored
+}
+
+func matchRule(r gitignorePattern, p string) bool {
+	if r.anchored {
+		// Match pattern against the whole path or any ancestor directory
+		// prefix — a rule like "build/" should ignore "build/foo" too.
+		if ok, _ := doublestar.Match(r.pattern, p); ok {
```
nenb, line 112: In general, we are silently dropping all these invalid globs (all the examples of `doublestar.Match`). I have found this frustrating in the past.

Is this the recommended approach?

internal/oci/walker.go
```diff
@@ -0,0 +1,264 @@
...
+			}
+			// gitignore dir-only patterns may elide a whole subtree.
+			if gitignoreMatches(gi, rel, true) {
+				return filepath.SkipDir
+			}
+			return nil
+		}
+
+		// Regular files only.
+		if !d.Type().IsRegular() {
```
nenb, line 230: GPT says that this is unknown on some (less common) platforms and environments and can lead to false negative. Suggests a refactor to something beginning like `d.Info().Mode().IsRegular()`.

At a minimum, think we should have a comment.

internal/oci/walker.go
```diff
@@ -0,0 +1,264 @@
+package oci
```
nenb, line 1: I think you know this already, but this file feels the most brittle to me in terms of how users could be surprised by Nebi's behaviour (e.g. how we determine what is ignored by git, how we parse the gitignore, how we perform the walk across all files).

Probably we should create a stub issue to track this i.e. identify footguns and either fix them or document them well?

cmd/nebi/import.go
```diff
@@ -44,19 +49,36 @@ func runImport(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("tag is required; use format registry/repository:tag (e.g., quay.io/nebari/my-env:v1)")
 	}
 
-	host, _ := oci.ParseRegistryURL(repoRef)
+	repoRef, plainHTTP := oci.StripScheme(repoRef)
 
 	ctx := context.Background()
```
nenb, line 56:
```suggestion
	ctx := cmd.Background()
```

cmd/nebi/import.go
```diff
@@ -88,7 +106,20 @@ func runImport(cmd *cobra.Command, args []string) error {
...
-	fmt.Fprintf(os.Stderr, "Imported %s -> %s\n", ref, absOutput)
+	fmt.Fprintf(os.Stderr, "Imported %s -> %s (%d asset file(s))\n", ref, absOutput, len(result.Assets))
 
 	return nil
 }
+
+// dirIsNonEmpty reports whether path exists and contains at least one
+// entry. Missing directory returns (false, nil).
+func dirIsNonEmpty(path string) (bool, error) {
+	entries, err := os.ReadDir(path)
```
nenb: Nit: For large dirs, this can be expensive. `File.ReadDir(1)` is cheaper?

cmd/nebi/import.go
```diff
@@ -44,19 +49,36 @@ func runImport(cmd *cobra.Command, args []string) error {
...
+	repoRef, plainHTTP := oci.StripScheme(repoRef)
 
 	ctx := context.Background()
-	result, err := oci.PullEnvironment(ctx, repoRef, tag, oci.BrowseOptions{
-		RegistryHost: host,
+
+	// Peek at manifest first so we can enforce the empty-destination
+	// policy before any bytes land on disk. This is cheap (one small
+	// GET) and avoids partial-extract state on a rejected destination.
+	peek, err := oci.PullBundle(ctx, repoRef, tag, oci.PullOptions{
```
nenb, line 61: This is another TOCTOU bug (potential). This and 74-80 and 93-96 can mention one tag and then extract from another.

It's probably low chance of occurring, and I don't think it can do much damage, but I think useful in preventing it all the same.

internal/oci/pathsafety.go
```diff
@@ -0,0 +1,115 @@
+package oci
```
nenb, line 1: Apologies about the following, but GPT has also flagged the following in this file, and I have not had time to verify. Please treat with some caution - verify first!

---------

### 1) Core-name collision check is case-sensitive (can bypass reserved core names)

 validateAssetPath only rejects exact "pixi.toml" / "pixi.lock" (internal/oci/pathsafety.go:84-87).
 So asset paths like Pixi.toml / PIXI.LOCK pass validation.

 validateAssetPaths collision logic (internal/oci/pathsafety.go:102-113) only detects collisions within
 assets, not between assets and the reserved core files written during extraction.

 Impact:
 - On case-insensitive filesystems (Windows, default macOS), an asset like Pixi.toml can collide
 with/overwrite the core pixi.toml path during extract.
 - This undermines the core-file layering invariant and can produce incorrect imported content.

 Suggested fix:
 - Apply normalizeForCollision to reserved core-name checks (not exact string equality), and/or
 - Seed collision detection with normalized reserved names (pixi.toml, pixi.lock) before processing asset
 paths.

---------

 ### 2) Windows-invalid filename chars are not fully blocked (: * ? " < > |)

 Current path safety checks block control chars, backslash separators, trailing dot/space, and reserved
 device names (internal/oci/pathsafety.go:35-50, 71-81), but do not reject several Windows-invalid
 characters in segments:
 < > : " | ? *

 Impact:
 - Bundles can carry asset names that are invalid or ambiguous on Windows.
 - : is especially important due to NTFS ADS semantics (e.g. file.txt:stream), which can create unexpected
 behavior/security issues.

 Suggested fix:
 - Reject Windows-invalid chars for each segment (or globally except / path separators).
 - Keep existing reserved device/trailing dot-space checks as additional guards.

nenb (commented):
Thanks! Very cool. 😎 

I have left a number of comments after review from GPT and myself.

I don't expect them all to be resolved before merging, but it would be nice to either have a resolution or an issue for each before merging.

internal/oci/publisher.go
aktech: fixed here https://github.com/nebari-dev/nebi/pull/295/changes/d2698550519e252865901e9adab6573e4e212771

internal/oci/publisher.go
aktech: Agreed, renamed to withAssets in https://github.com/nebari-dev/nebi/pull/295/commits/9042a1337dc502dbcbab777a5c707c39479e2fbe.

cmd/nebi/publish.go
aktech: Yeah agreed, that's not expected, fixed here [`de34470` (this PR)](https://github.com/nebari-dev/nebi/pull/295/commits/de3447087c6dff8930ac3722845855eaf286e729).

internal/oci/browser.go
aktech: Yep correct, fixed in https://github.com/nebari-dev/nebi/pull/295/commits/1a6eee4a25123ed5389f630873041cf0c04433b9.

internal/oci/browser.go
aktech: Yeah, I think that makes sense, fixed: https://github.com/nebari-dev/nebi/pull/295/commits/f39e03f76ac2169b4a89f8d39250951ff86b7bb8

internal/oci/browser.go
aktech: Fair point fixed here: https://github.com/nebari-dev/nebi/pull/295/commits/ad85f5e7cce0f5ac9915a3c05ff3ffdb39477e59.

internal/oci/browser.go
aktech: Fair, manifest + pixi files are always small in practice, but a capped read would be cheap defense. Creating a follow up issue.

internal/oci/walker.go
aktech: Yep, completely agree. We shouldn't pass errors silently, fixed here: https://github.com/nebari-dev/nebi/pull/295/changes/e0eb732ba83573ed7caa7e5508ebd1ac52e2430c

internal/oci/walker.go
aktech: I don't think the concern actually applies on modern Go, `newUnixDirent` detects that and automatically falls back to lstatat to populate the real type before the DirEntry is handed out.

see https://github.com/golang/go/blob/release-branch.go1.26/src/os/file_unix.go#L469-L489

internal/oci/walker.go
aktech: yep, I agree in fact I was hoping to find a library to do this, but didn't dig deeper in that route as all of this seems like someone would have already done it before, I have created an issue meanwhile to track this

cmd/nebi/import.go
aktech: Yep, fixed in https://github.com/nebari-dev/nebi/pull/295/commits/ad85f5e7cce0f5ac9915a3c05ff3ffdb39477e59

internal/oci/pathsafety.go
aktech: Verified and fixed (added test for each)

- https://github.com/nebari-dev/nebi/pull/295/commits/868fb2cffc2d6001ed03aa0a98125cbceda622d0
- https://github.com/nebari-dev/nebi/pull/295/commits/51a06de8fca1f8cc554993f189fbe3b90b73fdfa

internal/oci/publisher.go
aktech: Filed as follow-up: https://github.com/nebari-dev/nebi/issues/303

internal/oci/browser.go
aktech: https://github.com/nebari-dev/nebi/pull/295/commits/868fb2cffc2d6001ed03aa0a98125cbceda622d0
https://github.com/nebari-dev/nebi/pull/295/commits/cecae94018f9fd1c7a4192c2096fd8ee4594bdf9
https://github.com/nebari-dev/nebi/blob/de8d32faf89afa60658e8747734be716e44ddada/internal/oci/browser_test.go#L30-L55

internal/oci/walker.go
aktech: Yep makes sense to fix now, https://github.com/nebari-dev/nebi/pull/295/commits/7e2a621a8a677f689e3f2381ba8025510ae7ca82

internal/oci/walker.go
aktech: Added assertCoreFile (os.Lstat plus info.Mode().IsRegular()) in the publisher, running before loadBundleConfig so a symlinked pixi.toml is rejected with a clear "must be a regular file" error before anything is parsed or pushed.

Walker's force-include fallback switched to Lstat too for symmetry: https://github.com/nebari-dev/nebi/pull/295/commits/8150cc12d29172a0cad345b7f441d6c505d550b2

cmd/nebi/import.go
aktech: there is no such method

cmd/nebi/import.go
aktech: I think at this scale it wouldn't matter.

aktech: fixing, just in case

cmd/nebi/import.go
nenb: I don't follow.

I was proposing
```go
  ctx := cmd.Context()
  if ctx == nil {
      ctx = context.Background()
  }
```

At the moment, if the parent command cancels (`cmd`, the input to function), then this will still keep running i.e. won't be cancelled.

Issue #135, opened 2026-02-17
Add hierarchical namespaces for organizing workspaces with inherited permissions


Adam-D-Lewis:
## Problem

Current workspace organization is flat:
- All workspaces live at the same level
- No way to group related workspaces (e.g., all infra team workspaces)
- Permissions are per-workspace only
- To give someone access to 10 related workspaces, you must share each one individually

## Proposed Solution

Introduce **namespaces** as a separate concept from workspace names:

```
infra/proj-1:latest      # namespace=infra, workspace=proj-1, tag=latest
team-a/ml/exp-1:v2       # namespace=team-a/ml, workspace=exp-1, tag=v2
data-science:prod        # namespace=(root), workspace=data-science, tag=prod
```

Key features:
1. **Arbitrary depth**: `company/team/project/subproject/...`
2. **Permission inheritance**: Grant access at any level, inherits to children
3. **Separate from workspace name**: Namespace is organizational, name is identity

## New Commands

```bash
# Create namespace
nebi namespace create infra
nebi namespace create infra/prod

# Push to namespace
nebi push infra/my-workspace:v1
nebi push team-a/ml/experiment-1:latest

# Share at namespace level
nebi namespace share infra --user bob --role editor  # bob can access all workspaces under infra/*
nebi namespace share team-a/ml --team ml-team --role viewer

# List with namespace filtering
nebi workspace list --namespace infra
nebi workspace list --namespace team-a/ml
```

## Database Changes

- New `namespaces` table: `id`, `path`, `parent_id`, `name`, `owner_id`
- New `namespace_permissions` table: `user_id`, `namespace_id`, `role_id`, `inherit`
- Update `workspaces` table: Add `namespace_id` foreign key

## RBAC Changes

- Update Casbin model to support wildcard resource matching
- Resource format: `ns:infra/*` for namespace grants, `ws:{uuid}` for specific workspace
- Permission inheritance: check namespace hierarchy when evaluating access

## Local Namespaces

Namespaces aren't just for server-side organization - they also solve a local problem.

When importing from multiple sources (different OCI registries, different Nebi servers), workspace names can collide. Two orgs might both publish a workspace with `name = "data-science"` in their pixi.toml.

Apply the same namespace concept locally:

```bash
# Import with namespace prefix
nebi import quay.io/org-a/data-science:v1 --namespace org-a
nebi import quay.io/org-b/data-science:v1 --namespace org-b

# Reference by namespaced name
nebi shell org-a/data-science
[...]

PR #305, opened 2026-04-24
feat(config): add `default_packages` into every workspace

### LINKED ISSUE #288: [FEAT] -  Ability to define packages to be included in every environment
### Problem

When I use nebi environments in JLab, they need ipykernel (see #279 ). However, the failure mode when its not included is not visible to users. In this case, the environment builds in nebi fine, but when users select it as their notebook kernel, it thinks for a bit and then just "disconnects" from the notebook. The user has no visibility on the failure. 

This is one usecase, but there are others, e.g. a company wants to pin a minimum version of a package for all users for security reasons. 

### Proposed solution

Include the ability to dictate at the nebi admin level packages that will automatically be added to every nebi environment, even if the user doesn't request it. 

### Alternatives considered

_No response_

### Additional context

_No response_

viniciusdc:
## Summary

Closes #288

Adds a `default_packages` config field that lets admins declare packages automatically injected into every workspace at creation time. The primary motivation is ensuring `ipykernel` is always present so JupyterLab kernels don't silently disconnect — but it also covers security use cases like pinning a minimum package version org-wide.

Packages are injected at creation time only; users can remove them afterward via the normal package management flow.

### `internal/config/config.go`
- Added `DefaultPackages []string` to `PackageManagerConfig`

### `internal/executor/local.go`
- After both creation branches (custom `pixi.toml` path and `Init()` path), calls `pm.Install()` with `DefaultPackages` if any are configured

### `config.yaml`
- Documented the new field with `ipykernel` and `ipywidgets` as example values

## Before / After

```yaml
# Before — no way to inject packages into every workspace
package_manager:
  default_type: pixi

# After — admins can declare packages to inject at creation time
package_manager:
  default_type: pixi
  default_packages:
    - "ipykernel>=6"
    - "ipywidgets>=8.1,<9"
```


viniciusdc:
Right now I am not doing any changes to the toml in the UI view, if @marcelovilla PR for the dynamic read of the configuration comes first, we will not need to change anything here

Adam-D-Lewis:
Thanks for this work @viniciusdc. As I understand it, your motivation is to solve the case when a Nebari user wants to create a workspace, but forgets to include a kernel.  In that case, the workspace environment not usable from JupyterLab and there is no indication to the user what is going wrong.

Does https://github.com/nebari-dev/nb-nebi-kernels/issues/4#issuecomment-4360722447 solve the pain for your users?  It should be straight forward to add.  it instead gives a clear error message to users when they forget to add a kernel.

As far as enforcing certain package restrictions on workspaces, that is definitely something we want, but I think ideally that is implemented after [namespaces](https://github.com/nebari-dev/nebi/issues/135).  I think the model should be that namespaces are allowed to have conditions on them which are enforced when an workspace is created or updated.

Even for your motivating use case the PR as it stands isn't ideal b/c it forces a particular kernel (ipykernel), but it's valid to build an environment with a different kernel (e.g. r-kernel).  It seems like you'd want to be able to specify some more complex rules like workspace must have at least one of a set of packages (e.g. ipykernel, r-kernel, etc.), or better yet for your use case enforce that the package has a jupyter kernel of some type.

Some additional comments about this PR since we'll need to iron out these details at some point anyway:
- If a user pushes to Nebi via CLI, packages get added silently.  I favor rejection over silent addition. (e.g. Error: Push Rejected.  Workspace doesn't meet conditions: "Must include a jupyter kernel")
- This also silently adds packages when you import from an OCI registry. I'd prefer user being able to notice it.
- This currently only enforces packages on initial workspace creation and not on updates.  This means ipykernel could be removed.  Is that a feature or bug?
- Should we enforce these packages on `nebi init` and other local usage as well or no?

Adam-D-Lewis (changes requested):
https://github.com/nebari-dev/nebi/pull/305#issuecomment-4361651742

viniciusdc:
Hey @Adam-D-Lewis thanks for the nice reply, I think you absolutly right on the suggestions there, for the underlying issue of the missing ipykernel I think nb-nebi seems to be the best suited place for that and the most elegant solution in my mind as well -- one thing though, I would prefer if the install the dep message was overrdibel, so that we can point the user to use the nebi cli command instead of pixi to avoid drifting from local vs remote... (maybe that's not a possible scenario... not sure)

For the general enforcement of workspace policies like package requirements, I think your idea of the "namespaces" gateway is spot on and kind of reminds me of the Kubernetes approach to RBAC and enforcement policies. I don't have good answers for your questions there yet, though:

This currently only enforces packages during initial workspace creation, not on updates. This means ipykernel could be removed. Is that a feature or a bug?

This was intentional, to avoid too much burden on the implemntation and I talked with @aktech and we found this plausible initialy to let users chiose if they want to keep the dep or no... though for security situations, where some type of package is required as a security constraint I think the user should not be able to modify those envs... so maybe this could translate as a certain namespace permission role toggled from read/write...  not all ironed out yet...

Though I would like to close this PR in favor of the nb-nebi-kernel approach, it seems much more elegant and better suited for the actual program in hand right now.

dcmcand:
Closing in favor of @Adam-D-Lewis 's solution

PR #306, opened 2026-04-24
fix(worker): persist workspace path after creation

### LINKED ISSUE #294: Workspace path not stored in database, kernels not discoverable
## Summary

When creating a workspace via the Nebi UI in JupyterLab, the workspace files are created on disk at the expected path (`~/.local/share/nebi/workspaces/<name>-<uuid>/`), but the `path` column in the SQLite database (`nebi.db`) remains `NULL`. This causes:

1. `nebi workspace list --json` reports `"missing": true` for the workspace
2. `nb_nebi_kernels.NebiKernelSpecManager` skips missing workspaces
3. The workspace does not appear as a kernel in the JupyterLab launcher

## Steps to reproduce

1. Open JupyterLab with Nebi integration (via nebari-data-science-pack)
2. Go to the Nebi tab > Workspaces > New Workspace
3. Create a workspace (e.g., with `python = "3.9.*"` and `ray-serve = "2.43.*"`)
4. Wait for status to show "Ready"
5. Open the JupyterLab Launcher — the workspace kernel does not appear

## Diagnosis

```bash
$ nebi workspace list --json
[
  {
    "id": "1ba3ccd6-...",
    "name": "ray-serve",
    "status": "ready",
    "missing": true
    # no "path" field
  }
]

$ sqlite3 ~/.local/share/nebi/nebi.db "SELECT id, name, path FROM workspaces WHERE name='ray-serve';"
1ba3ccd6-...|ray-serve|

$ ls ~/.local/share/nebi/workspaces/
ray-serve-1ba3ccd6-d69b-41a1-a179-29ccbb411375/   # files exist on disk
```

## Workaround

Manually update the database:

```bash
sqlite3 ~/.local/share/nebi/nebi.db \
  "UPDATE workspaces SET path='$HOME/.local/share/nebi/workspaces/<name>-<uuid>' WHERE id='<uuid>';"
```

After this, `nebi workspace list` shows `"missing": false` and the kernel appears in the launcher.

## Expected behavior

The `path` column should be automatically populated when a workspace is created, since Nebi already knows where it writes the workspace files.

## Environment

- Nebi: v0.10.3
- nb_nebi_kernels: 0.1
- nebari-data-science-pack JupyterLab image: `sha-2044712`

viniciusdc:
## Summary

Closes #294

`SetWorkspacePath` was correctly writing the path to the DB, but `UpdateWorkspaceSize` runs immediately after and calls `db.Save(ws)` with the stale in-memory `ws` struct (loaded at job start with `path=""`). That full-row save was overwriting the path back to empty every time.

SQL evidence from the cluster:

```sql
-- SetWorkspacePath works:
UPDATE workspaces SET path="/home/jovyan/.local/share/nebi/workspaces/test8-54c8da39-..." WHERE id = "54c8da39-..."

-- UpdateWorkspaceSize immediately clobbers it:
UPDATE workspaces SET name="test8", path="", size_bytes=294733936, ... WHERE id = "54c8da39-..."
```

## Fix

After `SetWorkspacePath` succeeds, set `ws.Path = resolvedPath` in memory so the subsequent `db.Save(ws)` in `UpdateWorkspaceSize` preserves the path instead of overwriting it.

## How to test

Update `nebi.image.tag` in `clusters/hetzner-nebari/apps/nebi-pack.yaml` to the sha built from this branch, create a workspace via the UI, then verify:

```bash
export KUBECONFIG=~/Library/Caches/nic/hetzner-k3s/nic-nebari/kubeconfig
kubectl exec -n nebari-system <jlab-pod> -- sqlite3 /home/jovyan/.local/share/nebi/nebi.db \
  "SELECT name, path FROM workspaces ORDER BY created_at DESC LIMIT 5;"
```

The `path` column should be non-empty for newly created workspaces.

Keeping draft until verified on the cluster.


viniciusdc:
Tested on the hetzner cluster. After restarting the JLab pod with `sha-e81f609`, created `test11` and the path is correctly persisted:

```
test11 | /home/jovyan/.local/share/nebi/workspaces/test11-1a0a2608-70e7-4f20-88ab-e71772d658ba | ready
```

Earlier workspaces (`test8`, `test10`) still show empty path since they were created before the fix — expected.

internal/worker/worker.go
```diff
@@ -256,9 +256,16 @@ func (w *Worker) executeJob(ctx context.Context, job *models.Job, logWriter io.W
...
+		// Also update ws.Path in memory so the subsequent db.Save in UpdateWorkspaceSize
+		// does not overwrite the path back to "".
 		if ws.Path == "" {
-			w.svc.SetWorkspacePath(ws.ID, w.executor.GetWorkspacePath(ws))
+			resolvedPath := w.executor.GetWorkspacePath(ws)
+			if err := w.svc.SetWorkspacePath(ws.ID, resolvedPath); err != nil {
+				w.logger.Error("failed to persist workspace path", "workspace_id", ws.ID, "resolved_path", resolvedPath, "error", err)
+			} else {
+				ws.Path = resolvedPath
+			}
```
aktech, line 270: Can you write a regression test for this?

viniciusdc: Great catch, this is now included as part of  `internal/worker/worker_test.go`. Drives `executeJob` end-to-end with a stale `ws` via a minimal fake executor and asserts the `path` column is persisted.

viniciusdc:
@aktech the regression test is in and CI is green. Mind taking another look?

Adam-D-Lewis:
I started reviewing this but wanted to reproduce the error and found a new issue that prevented me from doing so. The linux nebi-desktop app crashes on startup.  https://github.com/nebari-dev/nebi/issues/350

viniciusdc:
Hey @Adam-D-Lewis, thanks for having a look, and good catch. I only ran this against NIC, so I wasn't aware that we needed to test this through the desktop version too. I wonder though, if testing on cloud would be enough as validation for this PR specifically, since the desktop probably would need to wait for upstream contribution or workaround to take effect first.

Adam-D-Lewis:
@viniciusdc Testing on cloud should be good enough to prove it works.  I was trying to reproduce locally to be sure I understand the issue.  With a recent fix, I was able to reproduce the issue today.  I plan on reviewing this tomorrow.  Thanks for your patience!

Adam-D-Lewis:
One concern on the error-handling branch: when `SetWorkspacePath` returns an error, the new code logs it but falls through — the create flow continues and `SetWorkspaceStatus(ws.ID, models.WsStatusReady)` still runs a few lines later. The user-observable result is identical to the bug this PR fixes: workspace shows `ready` in the UI but `path` is empty in the DB and the CLI reports `"missing": true`.

Suggest returning early (or transitioning to `WsStatusFailed`) after logging, matching the `CreateWorkspace` error handling four lines above:

```go
if err := w.svc.SetWorkspacePath(ws.ID, resolvedPath); err != nil {
    w.logger.Error("failed to persist workspace path", "workspace_id", ws.ID, "resolved_path", resolvedPath, "error", err)
    w.svc.SetWorkspaceStatus(ws.ID, models.WsStatusFailed)
    return err
}
ws.Path = resolvedPath
```

It's a low-likelihood path in practice (the only `SetWorkspacePath` failure mode is the DB write itself), but the whole point of the fix is to make the "ready with empty path" state unreachable — leaving the silent-continue branch keeps the original symptom possible.

Adam-D-Lewis:
On the regression test — nice that it drives `executeJob` end-to-end and actually reaches `db.Save(ws)` via the real `UpdateWorkspaceSize` path. One thing worth tightening:

```go
PackageManager: "test-pm-not-registered",
```

The test depends on `SyncPackagesFromWorkspace` and `CreateVersionSnapshot` failing with an unknown-manager error, and on the worker's current behavior of *logging and continuing* on those failures (`w.logger.Error("Failed to sync packages", ...)`). If either of those is later refactored to return early, fail-fast, or stop calling `UpdateWorkspaceSize` when the manager is invalid, this test stops covering the regression — and it does so silently, since it would still pass (the bug requires `db.Save(ws)` to actually run).

Two safer options:

1. Extend `fakeExecutor` (or add a sibling fake) to implement enough of the package-manager surface that `SyncPackagesFromWorkspace`/`CreateVersionSnapshot` succeed as no-ops on a registered manager name.
2. Use a registered manager (`"pixi"` or whatever the lightest one is) and inject test doubles for the two methods.

At minimum, a `// NOTE:` comment on the `PackageManager` line spelling out that the value is deliberately invalid and *why* the test still reaches the regression site would save the next person a debugging session.

Adam-D-Lewis (changes requested):
See comments

Adam-D-Lewis:
I also have a related follow-up PR if you have time to review @viniciusdc. https://github.com/nebari-dev/nebi/pull/353

Adam-D-Lewis (approved):
Both review points addressed and verified locally (vet clean, regression test passes):

1. **Error-handling branch** (`dc082123d`) — `SetWorkspacePath` failure now sets `WsStatusFailed` and returns instead of falling through to `ready`, making the "ready with empty path" state unreachable. 

2. **Regression test robustness** (`a4808dffb`) — registers a no-op package manager and writes empty `pixi.toml`/`pixi.lock` so the create flow stays on its happy path and reaches `db.Save(ws)` unconditionally, independent of future error-handling refactors. 

Thanks for the thorough fix and for verifying on the cluster. LGTM 🚀

PR #333, opened 2026-04-29
docs(cli-local): simplify bundle and import sections

khuyentran1401:
## Summary
Replace technical jargon (blob, opaque asset layers, bound, two-layer artifacts) in `docs/docs/cli-local.md` with plain language. Aimed at users who are not already familiar with OCI internals.

## Test plan
- [x] Render the docs site locally and verify the Publish/Import sections still read correctly

### DIFF
--- a/docs/docs/cli-local.md
+++ b/docs/docs/cli-local.md
@@ -95,30 +95,22 @@ nebi run ml-pipeline train -- --epochs 100
 
 ## Publish a Workspace Bundle
 
-`nebi publish --local` pushes the workspace directly to an OCI registry as a
-self-contained bundle. The artifact always contains `pixi.toml` and
-`pixi.lock`; any additional files in the workspace become opaque asset
-layers in the bundle.
+`nebi publish` packages your workspace and pushes it to an OCI registry.
+Every bundle includes `pixi.toml` and `pixi.lock`, plus any other
+workspace files (READMEs, source code, data) as additional layers.
 
 ```bash
-nebi publish --local --registry my-registry --tag v1
+nebi publish --registry my-registry --tag v1
 ```
 
 ### Selecting what goes into the bundle
 
-By default the walker includes every regular file in the workspace
-directory, with these rules applied in order:
+By default, the bundle includes everything in your workspace except
+`.git/` and `.pixi/`. `pixi.toml` and `pixi.lock` are always included
+no matter what.
 
-1. `.git/` and `.pixi/` are always dropped.
-2. If `[tool.nebi.bundle].include` is set in `pixi.toml`, only files
-   matching those globs are candidates.
-3. Files matched by `.gitignore` are dropped.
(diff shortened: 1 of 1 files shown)


docs/docs/cli-local.md
```diff
@@ -95,45 +95,42 @@ nebi run ml-pipeline train -- --epochs 100
...
-self-contained bundle. The artifact always contains `pixi.toml` and
-`pixi.lock`; any additional files in the workspace become opaque asset
-layers in the bundle.
+`nebi publish` packages your workspace and pushes it to an OCI registry.
+Every bundle includes `pixi.toml` and `pixi.lock`, plus any other
+workspace files (READMEs, source code, data) as additional layers.
 
 ```bash
-nebi publish --local --registry my-registry --tag v1
+nebi publish --registry my-registry --tag v1
```
khuyentran1401, line 103: `nebi publish` already defaults to local mode when no server is configured. Since that is the audience for this page, `--local` felt a bit redundant to me.

docs/docs/cli-local.md
```diff
@@ -95,45 +95,42 @@ nebi run ml-pipeline train -- --epochs 100
...
-`include` and `exclude` work together: `include` whitelists candidates,
-and `exclude` further narrows them down. When only `exclude` is set,
-everything is a candidate minus the excluded paths.
+- **`include`**: turn the default into a strict allowlist. Only files
+  matching these patterns are kept.
+- **`exclude`**: drop additional files (for example, `nebi.db*`) from
+  what's been kept so far.
+
+`.gitignore` rules also apply: files git ignores are kept out of
+bundles. Symlinks, device files, and named pipes are skipped silently.
```
khuyentran1401, line 127: The original numbered rules describe how the code applies filters, but as a user trying to drop something specific (like nebi.db*), it is hard to tell which rule to touch.

I reframed the section around what users start with and what they can change, and replaced technical terms like "globs," "candidates," and "FIFOs" with plainer wording.

docs/docs/cli-local.md
```diff
@@ -150,12 +147,16 @@ Tracking workspace 'data-science' at /home/user/my-project
 Imported quay.io/nebari/data-science:v1.0 -> /home/user/my-project (3 asset file(s))
 ```
 
-Use `--concurrency N` to tune parallel blob fetches (default 8).
+Use `--concurrency N` to set how many files download at the same time (default 8).
 
-When a bundle contains asset layers, the output directory must not
-already exist or must be empty — imports do not overwrite existing
-files. Legacy two-layer artifacts (no assets) keep the `--force` escape
```
khuyentran1401, line 157: I reworded this line because "legacy two-layer artifacts" and "escape hatch" are not terms most users would recognize.

pavithraes (approved):
Looks good to me, thanks!

PR #351, opened 2026-05-26
fix(desktop): re-apply SA_ONSTACK to survive WebKitGTK JSC signal handler clobbering

### LINKED ISSUE #350: [BUG] - nebi-desktop crashes on Linux/WebKitGTK (SA_ONSTACK fatal) since v0.10.5
### Description

`nebi-desktop` crashes on Linux (Ubuntu 26.04 + libwebkit2gtk-4.1 v2.52) shortly after launch with the Go runtime fatal:

```
Overriding existing handler for signal 10. Set JSC_SIGNAL_FOR_GC if you want WebKit to use a different signal
signal 11 received but handler not on signal stack
fatal error: non-Go code set up signal handler without SA_ONSTACK flag
```

Process exits with code 2. UI does not come up usefully — typically dies within ~1-2 seconds of the first frontend HTTP request landing.

### Steps to reproduce

1. On Ubuntu 26.04 (or any distro with libwebkit2gtk-4.1 ≥ 2.46), `pixi global install nebi==0.11`
2. Run `nebi-desktop`
3. Observe the crash within a few seconds.

Also reproduces from source:
```
git checkout v0.10.5  # or any later tag, or main
make build-desktop
./build/bin/Nebi
```

Does **not** reproduce on `v0.10.4`.

### Expected behavior

The desktop app comes up and stays running.

### Environment

```
OS:           Ubuntu 26.04 LTS (after recent dist-upgrade)
Webkit:       libwebkit2gtk-4.1-0 2.52.3-0ubuntu0.26.04.2 (also libwebkitgtk-6.0 present)
Nebi version: 0.10.5 through 0.11 (main)
Wails:        v2.11.0 (also reproduces on v2.12.0 — see below)
Go:           1.24
```

### Root cause

This is a **latent WebKitGTK/JavaScriptCore bug that Wails 2.x does not fully mitigate**. Nebi's `v0.10.5` simply removed a workload that was incidentally masking it.

**Mechanism:** WebKitGTK's JSC installs `SIGSEGV` handlers (used for generational-GC write barriers and thread suspension) **without `SA_ONSTACK`**, which Go's runtime refuses to tolerate. Wails 2.12 attempts to fix this by re-applying `SA_ONSTACK` once via `g_idle_add` after `gtk_init` — but JSC installs its handlers lazily on first JS context creation, often *after* that one-shot fix has already run. So Wails 2.12 alone is not sufficient.

What kept this latent before `v0.10.5`: Go's runtime calls `sigaction` (with `SA_ONSTACK`) every time it spawns a new OS thread to host a goroutine. Heavy startup work — DB queries, regex compilation, etc. — incidentally caused enough thread churn that `SA_ONSTACK` got re-applied on top of WebKit's bad handler before any JSC GC signal fired.

### Bisect

Bisected between `v0.10.4` (good) and `v0.10.5` (bad). 16 commits, 4 iterations. **First bad commit: `1546d97` — "test(e2e): bundle import via API round-trips assets in local mode".**

Despite the test-y name, the commit ships an `internal/api/router.go` change that skips `rbac.InitEnforcer` in local mode (`nebi-desktop` always runs in local mode):

```diff
-	if err := rbac.InitEnforcer(db, logger); err != nil {
-		logger.Error("Failed to initialize RBAC", "error", err)
-		panic(err)
+	if !cfg.IsLocalMode() {
+		if err := rbac.InitEnforcer(db, logger); err != nil {
+			logger.Error("Failed to initialize RBAC", "error", err)
+			panic(err)
+		}
 	}
```

Verification experiments at `1546d97`:

| Patch | Crash? |
|---|---|
| Revert just `router.go` (always call `InitEnforcer`) | **no** |
| Keep skip, replace with `time.Sleep(2 * time.Second)` | **yes** |

So it's not a nil-deref (`RequireAdmin` / `RequireWorkspaceAccess` middleware properly short-circuits in local mode and never touches the enforcer), and it's not pure wall-clock delay. It's specifically the OS-thread churn that `InitEnforcer`'s DB + casbin work induces.

### Proposed fix

Two options for nebi. Either alone is sufficient; I'd suggest doing both.

**1. Re-apply `SA_ONSTACK` on a short interval in the desktop binary.** Linux-only cgo, ~25 lines (a separate PR will follow). It also bumps `wails/v2` to `v2.12.0` (which improves but does not fix the case on its own — a Wails-side issue will follow with the same evidence).

```go
//go:build linux

package main

/*
#include <signal.h>

static void fix(int s) {
    struct sigaction st;
    if (sigaction(s, NULL, &st) < 0) return;
    if (!(st.sa_flags & SA_ONSTACK)) {
        st.sa_flags |= SA_ONSTACK;
        sigaction(s, &st, NULL);
    }
}

static void fix_all(void) {
    fix(SIGSEGV); fix(SIGBUS); fix(SIGFPE); fix(SIGILL); fix(SIGABRT);
}
*/
import "C"
import "time"

func init() {
    go func() {
        t := time.NewTicker(50 * time.Millisecond)
        defer t.Stop()
        for range t.C { C.fix_all() }
    }()
}
```

**2. Track and pick up the upstream Wails fix.** Wails 2.12's `g_idle_add` is one-shot; a recurring `g_timeout_add` or a hook on the WebView's `load-changed` signal would resolve the underlying race. A Wails-side issue will be filed with the same evidence.

### Additional context

- `nebi serve` + browser remains a working alternative in the meantime (the embedded HTTP path doesn't involve WebKit at all).
- The same crash signature appears in Wails issues [#2134](https://github.com/wailsapp/wails/issues/2134) and [#3965](https://github.com/wailsapp/wails/issues/3965). Wails 2.12.0's changelog says #3965 was fixed, but our empirical evidence shows the fix is insufficient for cases where the Go runtime doesn't churn enough OS threads at startup.

Adam-D-Lewis:
Fixes #350.

WebKitGTK's JavaScriptCore installs SIGSEGV/SIGBUS/SIGUSR1 handlers (for
GC thread suspension, write barriers, Wasm trap handling) without
SA_ONSTACK, which Go's runtime refuses to tolerate. Wails 2.12.0 added
a one-shot mitigation but JSC installs its handlers lazily on first JS
context creation — usually after that one-shot fix has already run.

This PR:
- bumps `wails/v2` from 2.11.0 → 2.12.0
- adds `signal_fix_linux.go`: a 50ms cgo ticker that re-applies SA_ONSTACK,
  doing explicitly what Go's runtime was doing accidentally via OS-thread
  `sigaction` calls (see issue #350 for the bisect that uncovered this)

Verified on Ubuntu 26.04 LTS + libwebkit2gtk-4.1 v2.52.3.

## Suggested follow-up: release after merge

The broken 0.10.5 and 0.11 linux-64 builds are being marked broken on
conda-forge via [conda-forge/admin-requests#2104](https://github.com/conda-forge/admin-requests/pull/2104).
Once that lands, `pixi global install nebi` on linux-64 will resolve to
0.10.4 (the last known-good release). Once this fix PR is merged we
should cut a new release (suggest **0.11.1** or **0.12**) so linux-64
users get a current `nebi-desktop` back rather than being pinned to
0.10.4 indefinitely.

## Removal of the workaround

Once [wailsapp/wails#5507](https://github.com/wailsapp/wails/pull/5507)
ships in a tagged Wails release and we bump to it, `signal_fix_linux.go`
can be deleted (the comment in that file documents this).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/swaggo/gin-swagger v1.6.1
 	github.com/swaggo/swag v1.16.6
 	github.com/valkey-io/valkey-go v1.0.67
-	github.com/wailsapp/wails/v2 v2.11.0
+	github.com/wailsapp/wails/v2 v2.12.0
 	github.com/zalando/go-keyring v0.2.8
 	golang.org/x/crypto v0.43.0
 	golang.org/x/oauth2 v0.33.0
@@ -34,6 +34,7 @@ require (
 )
 
 require (
+	git.sr.ht/~jackmordaunt/go-toast/v2 v2.0.3 // indirect
 	github.com/KyleBanks/depth v1.2.1 // indirect
 	github.com/bep/debounce v1.2.1 // indirect
 	github.com/bytedance/gopkg v0.1.3 // indirect
--- a/go.sum
+++ b/go.sum
@@ -1,3 +1,5 @@
+git.sr.ht/~jackmordaunt/go-toast/v2 v2.0.3 h1:N3IGoHHp9pb6mj1cbXbuaSXV/UMKwmbKLf53nQmtqMA=
+git.sr.ht/~jackmordaunt/go-toast/v2 v2.0.3/go.mod h1:QtOLZGz8olr4qH2vWK0QH0w0O4T9fEIjMuWpKUsH7nc=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/KyleBanks/depth v1.2.1 h1:5h8fQADFrWtarTdtDudMmGsC7GPbOAu6RVB3ffsVFHc=
 github.com/KyleBanks/depth v1.2.1/go.mod h1:jzSb9d0L43HxTQfT+oSA1EEp2q+ne2uh6XgeJcm8brE=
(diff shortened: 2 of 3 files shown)


viniciusdc (approved):
Nice work, looked at the change good catch. I just ask to open a follow up issue after this gets merged so that we have a tracking issue open on our side as well to remove this once that PR lands.

PR #356, opened 2026-05-28
docs(home): show CLI before Web UI in demo section

khuyentran1401:
## Reference Issues or PRs

N/A

- Swaps the order of the "See Nebi in Action" demos on the home page so the CLI appears first and the Web UI second.
- Leads with the CLI, which is the primary entry point for most users, before showing the Web UI experience.

## Testing

- [ ] Did you add new tests?

Verified the new order renders correctly on the local Docusaurus dev server.

### DIFF
--- a/docs/src/pages/index.tsx
+++ b/docs/src/pages/index.tsx
@@ -42,19 +42,19 @@ function HomepageDemo() {
         </Heading>
         <div className={styles.demoGrid}>
           <div className={styles.demoItem}>
-            <Heading as="h3" className={styles.demoLabel}>Web UI</Heading>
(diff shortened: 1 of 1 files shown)


smeragoel:
Thanks @khuyentran1401 for the quick turnaround on this! I think it looks great. I have a small nit, the two gifs have different vertical heights, would it be possible to make them the same height by resizing the CLI gif? 

image

This issue existed before this PR, I just thought it might be good to fix that too!

(Sorry for the PR close, I accidentally hit some shortcut that pasted my comment prematurely and closed the PR)

jbouder (approved):

PR #386, opened 2026-06-05
fix(frontend): show group shares in Collaborators tab

aktech:
## Problem

Sharing a workspace with a group added the group to the backend grants and showed it in the Share dialog's "Current Access" list, but the **Collaborators tab** on the workspace detail page rendered nothing for that group. The tab fetched both user and group grants, then only mapped over the user list, so groups were silently dropped (and the tab count excluded them).

## Fix

- Add a read-only `CollaboratorsList` component that renders both users and groups. Group rows show the group name, its source (`Native group` / `OIDC group`), and the role badge, matching how the Share dialog already presents groups.
- Use it in the Collaborators tab in place of the user-only map.
- Base the tab count on the full collaborator list instead of just users.

Backend (`GET /workspaces/{id}/collaborators`) already returned groups, so no server change was needed.

## Tests

New `CollaboratorsList.test.tsx`: renders user rows, renders group rows alongside users, distinguishes native vs OIDC group labels, and shows the empty state.

### DIFF
--- a/frontend/src/components/sharing/CollaboratorsList.test.tsx
+++ b/frontend/src/components/sharing/CollaboratorsList.test.tsx
@@ -1,0 +1,54 @@
+import { render, screen } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+import type { Collaborator } from '@/types/models';
+import { CollaboratorsList } from './CollaboratorsList';
+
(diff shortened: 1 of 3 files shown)


jbouder (approved):
Works and looks great!

PR #387, opened 2026-06-05
ci: inject VERSION build-arg into docker image build

aktech:
Published nebi images report their version as `dev`, so the web UI shows **`vdev`**.

## Root cause

The runtime chain is correct: `-X main.Version` -> `serve.go` -> `server.Config.Version` -> `handlers.Version` -> `/version`. The Dockerfile already accepts `ARG VERSION` and bakes it via ldflags.

The break is in CI: the `Build and maybe push` step in `docker.yml` never passed a `VERSION` build-arg, so `ARG VERSION=dev` (the Dockerfile default) was used. The container ships no `.git`, so the handler's runtime `git describe` fallback also fails, leaving `version: "dev"`.

## Fix

- Compute the version with `git describe --tags --always --dirty` (the exact format `internal/api/handlers/version.go` parses, and what the `Makefile` already uses).
- Pass it as `build-args: VERSION=...` to `docker/build-push-action`.
- Add a smoke-test guard that fails the build if `/version` reports `dev`.

## Verification

Built the image locally with the build-arg and ran the smoke test:

```json
{ "version": "0.12-rc3.dev+e8f6759", "commit": "e8f6759", ... }
```

Previously this returned `"version": "dev"`.

Screenshot 2026-06-05 at 13 41 55

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -35,6 +35,14 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
 
+      - name: Compute version
+        id: version
+        # git describe matches the format internal/api/handlers/version.go parses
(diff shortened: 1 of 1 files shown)


viniciusdc (approved):
LGTM, the test addition trough the api request was a clever idea

PR #404, opened 2026-06-10
make workspace name optional

### LINKED ISSUE #216: Bug: Workspace name ignored on creation — defaults to 'my-project'
## Summary

When creating a workspace via the GUI, the user-entered workspace name is ignored in the pixi.toml. The workspace is created with `name = "my-project"` regardless of what name was entered.

## Steps to Reproduce

1. Open the desktop app (local mode)
2. Click "Create Workspace"
3. Enter a workspace name (e.g., "my-data-project")
4. Click "Create Workspace"
5. Observe the created workspace's pixi.toml has `name = "my-project"`

## Root Cause

The default pixi.toml template in `frontend/src/pages/Workspaces.tsx` has a hardcoded name:

```toml
[workspace]
name = "my-project"
```

The `newWsName` state (user input) and `pixiToml` state (TOML content) are independent and never synced. The `PixiTomlEditor` component receives `workspaceName` as a prop but only uses it when switching between UI/TOML editing modes — not on initial load or form submission.

The backend executor (`internal/executor/local.go`) writes custom pixi.toml content as-is without updating the name field to match the workspace record.

## Relevant Code

- `frontend/src/pages/Workspaces.tsx` — `DEFAULT_PIXI_TOML` (hardcoded name), form submission
- `frontend/src/components/workspace/PixiTomlEditor.tsx` — `workspaceName` prop unused on mount
- `internal/executor/local.go` — writes pixi.toml without correcting name

## Possible Fixes

- **Frontend**: Sync the pixi.toml name field whenever the workspace name input changes (e.g., `useEffect` on `workspaceName` prop in `PixiTomlEditor`)
- **Backend**: Have the executor replace the `[workspace] name` in pixi.toml with the workspace's actual name before writing it

### LINKED ISSUE #324: [BUG] -  Local workspace name is being overridden
### Description

I created a workspace in the UI, and added a workspace name in the UI box. The workspace name in the toml remained "my-project". The environment successfully created, but then a few seconds later, it renamed to "my-project". 

I did not see the same behavior for remote environments. 

Also, because of [this bug](https://github.com/nebari-dev/nebi/issues/294) I am defining the workspace path as well. The path did not change.  

### Steps to reproduce

1. Open Nebi via JLab
2. Click "local"
3. Click "New workspace"
4. Add a workspace name and path via the UI boxes. Leave the toml as the default:

Image

5. Click "create and save"

Image

6. Go back to Workspaces

Image

7. wait maybe 30 seconds, spend your time clicking around on different things on nebi... 
8. Go back to Workspaces and witness your workspace has been renamed

Image

Image

### Expected behavior

I expect the workspace name I set to persist. 

### Environment

```
    repoURL: https://github.com/nebari-dev/nebari-nebi-pack.git
    targetRevision: 7a77b96
    path: .
    helm:
      releaseName: nebi-pack
      values: |
        image:
          tag: sha-e81f609
```

### Additional context

I will also mention that the duplication of workspace name in the UI AND the toml default is a little confusing anyway. I suspect the solution here should be to ensure that name defined in the UI text box should override what's in the toml and that we should remove that field from our default toml.

### LINKED ISSUE #348: [FEAT] -  Remove UI widget for workspace name
### Problem

Currently, we have a UI widget for users to input the workspace name. The pixi.toml also enables users to add a workspace name. And in fact, in the default example pixi.toml we display to users, there is a workspace name included. Which one will nebi use? Users are left to guess. 

In addition to confusing users, it seems the backend is also confused about this over-specification. See #324 and #216 . 

It would be best if workspace name were only specified in one location. 

### Proposed solution

My vote would be the pixi.toml to allow for our users to reuse a single pixi.toml in or out of nebi. 

### Alternatives considered

_No response_

### Additional context

_No response_

pmeier:
## Reference Issues or PRs

Fixes #216, fixes #324, and closes #348. 

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

#### Text styling

#### Non-text content

### DIFF
--- a/frontend/src/pages/Workspaces.tsx
+++ b/frontend/src/pages/Workspaces.tsx
@@ -56,6 +56,7 @@ platforms = ["osx-arm64", "linux-64"]
 python = ">=3.11"
 `;
 
+// TODO: Robustify. Maybe use a proper TOML parser?
 const getTomlName = (toml: string): string | null => {
   const match = toml.match(/^name\s*=\s*"([^"]*)"/m);
   return match ? match[1] : null;
--- a/frontend/src/pages/Workspaces.tsx
+++ b/frontend/src/pages/Workspaces.tsx
@@ -56,6 +56,11 @@ platforms = ["osx-arm64", "linux-64"]
 python = ">=3.11"
 `;
 
+const getTomlName = (toml: string): string | null => {
+  const match = toml.match(/^name\s*=\s*"([^"]*)"/m);
+  return match ? match[1] : null;
+};
+
 export const Workspaces = () => {
   const navigate = useNavigate();
   const setPendingTab = useWorkspaceNavStore((s) => s.setPendingTab);
@@ -73,7 +78,6 @@ export const Workspaces = () => {
 
   const [showCreate, setShowCreate] = useState(false);
   const [createTarget, setCreateTarget] = useState<'local' | 'server'>('local');
-  const [newWsName, setNewWsName] = useState('');
   const [localPath, setLocalPath] = useState('');
   const [pixiToml, setPixiToml] = useState(DEFAULT_PIXI_TOML);
 
@@ -137,7 +141,11 @@ export const Workspaces = () => {
 
   const handleCreate = async (e: React.FormEvent) => {
     e.preventDefault();
-    if (!newWsName.trim()) return;
+    const wsName = getTomlName(pixiToml);
+    if (!wsName || !wsName.trim()) {
+      setError('Workspace name is required in the pixi.toml [workspace] section.');
+      return;
+    }
 
     setError('');
     try {
@@ -145,13 +153,13 @@ export const Workspaces = () => {
 
       if (createTarget === 'server' && isRemoteConnected) {
         await createRemoteMutation.mutateAsync({
-          name: newWsName,
+          name: wsName,
           package_manager: 'pixi',
           pixi_toml: tomlContent,
         });
       } else {
         const ws = await createMutation.mutateAsync({
-          name: newWsName,
+          name: wsName,
           package_manager: 'pixi',
           pixi_toml: tomlContent,
           ...(localPath.trim()
@@ -160,7 +168,6 @@ export const Workspaces = () => {
         });
 
         // Reset form
-        setNewWsName('');
         setLocalPath('');
         setPixiToml(DEFAULT_PIXI_TOML);
         setShowCreate(false);
@@ -171,7 +178,6 @@ export const Workspaces = () => {
       }
 
       // Reset form
-      setNewWsName('');
       setLocalPath('');
       setPixiToml(DEFAULT_PIXI_TOML);
       setShowCreate(false);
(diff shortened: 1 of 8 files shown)


pmeier:
## TOML mode

image

image

## UI mode

image

image

viniciusdc:
@jbouder can you have a look?

internal/service/workspace.go
```diff
@@ -116,8 +117,20 @@ func (s *WorkspaceService) Create(ctx context.Context, req CreateRequest, userID
 		packageManager = "pixi"
 	}
 
+	name := req.Name
+	if name == "" && req.PixiToml != "" {
```
pmeier: maybe create a ResolveWorkspaceName helper in the pixi namespace to avoid duplication below.

frontend/src/pages/Workspaces.tsx
```diff
@@ -56,6 +56,11 @@ platforms = ["osx-arm64", "linux-64"]
 python = ">=3.11"
 `;
 
+const getTomlName = (toml: string): string | null => {
```
pmeier, line 60: This is not particularly robust. We can do that later. I'll add a comment.

jbouder:
I'm still seeing my-project when i create a new workspace. Just need to fix that.
Screenshot 2026-06-10 at 12 00 44 PM

pmeier:
@jbouder fixed in 1da10634b8e204a1ee493b01ef4214bab47c3fef

image

smeragoel:
Thanks @pmeier for working on this!

I took a UX pass over the resulting form and put together some suggested adjustments in Figma: [Create Workspace — suggested layout](https://www.figma.com/design/Wu8VTFA5IO5BjbvifZJMyZ/Nebi?node-id=1072-1666&t=Z3XPNxbiPodlpnuL-11) (four frames: TOML mode initial state, TOML mode filled, UI mode, Edit Configuration).

> Note: the mocks are built with the new Nebari Design System because that's what my working file uses. Everything below applies to the current UI as-is for the structure/behavior, not the styling. The NDS visual style is not part of this suggestion. 

1. Reorder the form -- mode toggle first, Path last
image

With the Name field removed, Path (optional) becomes the first input but not intentionally. It's an optional field and leading with it signals it's required and makes "where" come before "what." Suggested order: mode toggle → content (toml editor / name + packages) → Path at the bottom.

2. Add a derived name line to make the name field more visible.
image

The name now lives inside the toml, which is correct but can be easily missed. Nothing tells users that line 2 of the editor is where the workspace gets its name, and the "name is required" will appear only on submit. Suggestion: a live line directly under the editor  (`Workspace will be created as: my-data-project`) that updates as you type. Create workspace also remains disabled until the user provides the name.

3. Don't prefill a real name in the default toml
This is to make sure that everyone intentionally set the workspace name rather than accidentally creating duplicates because they missed the name field.

pmeier:
Thanks @smeragoel for the review. @jbouder do you want to have the suggested changes in this PR or in a follow-up PR?

jbouder (approved):

PR #418, opened 2026-06-24
only poll remote server from frontend in local mode

pmeier:
Regression from #93. Without this guard, the frontend unconditionally polls the `/api/v1/remote/server` endpoint, which only exists in local mode.

### DIFF
--- a/frontend/src/hooks/useRemote.test.ts
+++ b/frontend/src/hooks/useRemote.test.ts
@@ -1,6 +1,7 @@
 import { renderHook, waitFor } from '@testing-library/react';
 import { HttpResponse, http } from 'msw';
 import { beforeEach, describe, expect, it } from 'vitest';
+import { useModeStore } from '@/store/modeStore';
 import {
   mockJob,
   mockRegistry,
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/useRemote.test.ts
```diff
@@ -35,6 +36,7 @@ const mockRemoteWorkspace = {
 
 describe('useRemoteServer', () => {
   beforeEach(() => {
+    useModeStore.setState({ mode: 'local' });
```
Adam-D-Lewis: @pmeier Not causing a problem at the moment, but the value leaks into the tests that run after this block. Probably worth setting `useModeStore` state back to null in an `afterEach`.

Adam-D-Lewis: I'll go ahead and approve as is though.

Adam-D-Lewis (approved):

tylerpotts:
Rebased onto main (the useRemote.ts hook file was heavily reworked by https://github.com/nebari-dev/nebi/pull/506 and others, so it conflicted). Kept the `enabled: isLocalMode` guard, and while resolving I also addressed the review nit: the test now resets `useModeStore` in an `afterEach` and adds a case asserting the query stays idle in team mode. Verified the bug is still live on main: `useRemoteView` in Layout polls `/api/v1/remote/server` every 10s regardless of mode, and the backend only registers the `/remote` group when `localMode` is true.

PR #419, opened 2026-06-24
fix SSE for multiline pixi logs

pmeier:
Previously we blindly wrote `"data: %s\n\n"`. If `%s` contains a linebreak `\n`, the event reader on the client terminated the message to early and discarded the remainder because it was not prefixed with `data: `. The result was that during workspace creation sometimes log lines were missing that should have been streamed in. The same lines were available after page refresh, because the data got lost in the transfer from server to client rather than on the client.

### DIFF
--- a/internal/api/handlers/job.go
+++ b/internal/api/handlers/job.go
@@ -3,6 +3,7 @@ package handlers
 import (
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/gin-gonic/gin"
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


tylerpotts:
Thanks for tracking this down. The diagnosis was right, and I'm closing this in favor of https://github.com/nebari-dev/nebi/pull/536 for two reasons:

1. **The SSE framing fix already landed.** https://github.com/nebari-dev/nebi/pull/466 added `writeSSEEvent`/`writeSSEData` in `internal/api/handlers/job.go`, which splits multiline chunks so every line gets its own `data:` prefix, and additionally normalizes `\r\n` and `\r`. When I resolved this branch's conflicts against `main`, the entire `job.go` diff collapsed to nothing; the only surviving change was the broker buffer bump in `internal/logstream/broker.go`.

2. **The buffer bump alone still drops lines silently.** `LogBroker.Publish` does a non-blocking send and discards the line when the subscriber's buffer is full. Going from 100 to 4096 raises the threshold but a large pixi solve against a slow client still loses lines with no indication, which reproduces exactly the symptom described here (gaps in the live stream, complete logs after refresh), just less often. #536 keeps the 4096 buffer and adds a bounded 100ms wait before dropping, plus a `[nebi] N log lines dropped ...` marker injected into the stream so any remaining gap is visible instead of silent.

#536 also adds tests for both failure modes: a browser-style SSE parser driving the real `streamLogsFromBroker` loop with a burst of multiline chunks, mutation-checked to fail against both the old `data: %s\n\n` format and the old 100-entry buffer.

PR #420, opened 2026-06-24
improve check for existing workspaces

pmeier:
Instead of checking if just the workspace folder exists, we are now checking whether the manifest and the lock file exist in that folder. If they don't, the workspace can be pruned and is not listed.


tylerpotts:
Rebased this onto current main (no conflicts, 7b2d94b -> 7b16b78) and added a small table-driven unit test for workspaceDirMissing in cmd/nebi/workspace_test.go (1e3d627). It covers: nonexistent dir, empty dir, unrelated files only (all treated as missing), and manifest only / lockfile only / both (all treated as present).

tylerpotts (approved):
Approving. Can merge when CI is green

PR #421, opened 2026-06-24
allow new jobs if the workspace is in a failed state

pmeier:


tylerpotts:
Closing in favor of https://github.com/nebari-dev/nebi/pull/521, which fixes https://github.com/nebari-dev/nebi/issues/497 the same place this PR identified (the status gates in `submitJob` and `RollbackToVersion`) but with a smaller change that fits current `main`.

Two things changed since this branch was written that made us go a different way:

1. `main` has since gained the job-admission rework: `submitJob` now runs in a transaction with `lockJobAdmission` and per-user/per-workspace/global active-job quotas (`checkActiveJobQuotas`). That's the source of the merge conflict, and it already covers the concurrent-job protection `hasActiveJob` was providing here.
2. On top of that machinery, `hasActiveJob` would effectively hard-cap active jobs at one per workspace, overriding the configurable `ActiveJobsPerWorkspace` limit (default 2), and several quota tests would need rewriting to accommodate that. We didn't want to add a new limitation as a side effect of the bug fix.

#521 instead just relaxes the status gates to accept `failed` alongside `ready` (transitional states still reject, with the actual status now included in the error message), keeping concurrency entirely under the existing quotas. Verified live: a workspace bricked by a bad pin recovers with a corrected spec on the next solve, including workspaces already stuck in `failed` from before the fix.

Thanks @pmeier — this PR pinpointed the right fix location and its test-helper insight about lingering create jobs saved time when evaluating the approaches.

Issue #413, opened 2026-06-15
[FEAT] -  Separate TOML and UI editing instead of a bidirectional toggle


smeragoel:
### Problem

We've had a run of bugs and PRs this week around creating and editing workspaces ( #411, #410 + more). We keep patching the individual leaks, but the create/edit experience has become fragmented because we're treating the symptoms, not the cause.

The cause is the bidirectional toggle between UI mode and TOML mode. This bidirectional sync assumes the two are equivalent representations but they are not. pixi.toml is a superset of UI mode. It can represent things that the UI mode cannot (tasks, comments, custom channels). And this set isn't fixed, so we can't also dynamically generate UI fields for whatever's in the TOML.

So the two directions aren't symmetric:
- UI → TOML is lossless (the form is a subset)
- TOML → UI is lossy (anything UI mode can't represent gets dropped)

A bidirectional toggle forces the lossy direction to happen which is the root of the data-loss and discard bugs. 

### Proposed solution

I suggest separating the two experiences and only ever allow the lossless direction (UI → TOML), never the lossy one.

#### Create

1. "New Workspace" opens the GUI editor directly without any toggle.

2. The split-button dropdown adds "Create from pixi.toml", which drops you straight into the TOML editor.
<img width="2900" height="1872" alt="Image" src="https://github.com/user-attachments/assets/38b43ac3-efbd-45a3-80fd-3dfae97dc5f2" />

3. From the UI editor, if you realize you need custom config the form doesn't support, you can go to the TOML editor which is fine, because UI → TOML is lossless and all your entered settings carry over.

<img width="2900" height="1872" alt="Image" src="https://github.com/user-attachments/assets/460e4065-a9b1-450e-ba0b-0695145b7ba8" />

4.There is no way to switch back quickly. Going back means abandoning the form and its data, and that's an expected UX pattern: leaving one form to start a different one loses your input. It's a completely different form, not a toggle, so the expectation is correct.

<img width="2900" height="1872" alt="Image" src="https://github.com/user-attachments/assets/6d1d7608-9941-441d-a778-7a4549c449a0" />

5. Advanced users can set their default create mode so they skip the dropdown each time.

#### Edit

1. Move the global Edit button to a per-tab Edit.
2. Edit on the Packages tab = UI-mode editing, scoped to packages. Saving goes UI → TOML: it updates only the dependencies and doesn't touch anything else in the file (name, channels, tasks, etc.).

[...]

Issue #327, opened 2026-04-28
[ENH] Provide feedback after "Save & Install"


pavithraes:
## Description

After editing a workspace, and clicking the "save & install", we just go back to the "Configuration" page.

It'll be helpful to have give feedback to the user, saying something like "A job has started, click here for logs". It can potentially be some kind of pop up message.

The link to logs is useful to track status.

## Next steps

Define the UI/UX here, and implement it

_Note: This was originally noted by @kcpevey in a user workflow discussion._

kcpevey:
I also find it confusing that the save and install button is not in the same place as the "create workspace" on the bottom right.

kcpevey:
The core problem here is likely that the "Ready" pill on the environment doesn't update. When I click the save button, the UI doesn't show the "Running" pill like it should.

PR #425, opened 2026-06-26
user stories for roadmap development

pmeier:
nebi has grown organically since its inception in a reactionist environment without a clear roadmap. This PR adds user stories that will drive the roadmap development.


smeragoel:
Went through the user stories doc, it's in a really good shape. I really like the glossary especially! A few suggestions:

1. The user stories slice personas by deployment context (solo / team / automation / cluster), but I don't know where to fit the notebook user (the data scientist on Nebari who just picks a Nebi environment as a Jupyter kernel and runs a notebook). We have a lot of kernel bugs (#294, the launcher work), should we either add a stakeholder for them or add kernel stories into an existing persona?

2. There are no stories for editing an environment. Is that intentionally omitted because it's a basic story? A big chunk of what we've been designing (the save-and-install feedback work in #327, the TOML/UI editing split in #413) is based on this story.

3. There are currently no failure / error-handling stories. Something like "As a developer, when an environment fails to build, I want to understand why and be assured my previous working version is intact

dharhas:
Agree with @smeragoel, this might be missing a persona. 

I think the `developer` persona kinda needs to be split into two similar but different personas

A developer is expected to actively mutate an environment, they create, pull, push, modify environments and are actively developing software.

A secondary use case is a `consumer/data scientist/analyst/ai-user/etc` not sure what the best term is. They are going to pull a authorized or vetted environment and just use it. Either to write & run jupyter notebooks or run a packaged software that is distributed via nebi. They are unlikely to edit or modify environments but might need to update to the latest published environment or rollback to an earlier one etc.

pavithraes:
I'll expand on Dharhas' comments that we can include _using the workspace/environment_ as something developers would want to do (also where `nebi shell` & `nebi run` will fit in)

More broadly, it might be worth considering and documenting higher-level core motivation/goal for each of these stakeholders. Right now we have several tactical needs, a higher-level motivation could look like "I want my environments to be easy to update with version control, so that I can iterate on my projects quickly"

PR #440, opened 2026-07-10
Explicit workspace installation in local mode

### LINKED ISSUE #438: Explicit Workspace Installation in Local Mode
## Summary

The nebi server stops running `pixi install` during workspace creation and environment solving. Instead it runs `pixi lock`, which produces a lockfile without downloading or extracting packages. In local mode (`NEBI_MODE=local` env var, not the `--mode` serve flag), installing the environment becomes an explicit user action: each workspace shows an install status indicator and an install button in the UI, and the CLI gains `nebi workspace install` plus a `--installed` filter on `workspace list`. This eliminates the confusing implicit install that currently fires when a user first tries to use a shared workspace, which for large environments looks like a hung process.

## Goals

- Server-side workspace creation and solving produce only `pixi.lock` — no `.pixi/envs/` directory, no package downloads.
- In local mode, the install step is explicit: the user clicks a button or runs a CLI command. Workspaces default to locked-but-not-installed.
- The CLI `workspace list` command gains a `--installed` flag and indicates install status in its default output.
- Rollback and manifest updates auto-install if the workspace was installed before the operation, keeping the environment in sync with the latest lockfile.
- Explicit uninstall: `nebi workspace uninstall <name>` removes `.pixi/envs/`. The UI toggles between Install and Uninstall based on current state.

## Non-Goals

- Cleaning up existing `.pixi/envs/` directories on team-mode (`NEBI_MODE=team`) servers that accumulated dead-weight environments. Manual cleanup only; those servers never needed installed environments (OCI publish already operates from `pixi.toml` + `pixi.lock`).
- Changes to uv workspaces. uv support is dead code and will be removed separately.

## Background / Motivation

**Current behavior.** When a workspace is created or its manifest is updated, the server runs `pixi install`. This downloads and extracts all packages into `.pixi/envs/` — often gigabytes of data. Package metadata is then synced into the database via `pixi list` against the installed environment, and the on-disk size is measured for the UI.

In local mode (`NEBI_MODE=local`), which is the current proxy for a proper nebi client, the UX is confusing: when a workspace is shared from a remote, it appears in the UI but the environment is not installed. Attempting to use it triggers a lazy implicit install with no progress indication. For large workspaces this looks like broken behavior — nothing appears to happen while packages download in the background.

**Proposed behavior.** The server never installs environments. It runs `pixi lock`, which resolves the full dependency graph and writes `pixi.lock` (~15 KB for a typical workspace vs. gigabytes for the installed env). The lockfile is the source of truth for package metadata. Installing becomes an explicit user action in local mode, reusing the existing async job pipeline with progress streaming.

This aligns with how the OCI publish path already works: `PublishPixiOnly` pushes only `pixi.toml` and `pixi.lock` — the installed environment was never part of the artifact. The server was installing packages solely to list them and measure size. Listing already works from the lockfile; size moves to per-machine environment metadata.

### Server-side: `pixi lock` instead of `pixi install` (all modes)

Two code paths change:

1. **Workspace creation** (`executor.CreateWorkspace` / `worker.processJob` for `JobTypeCreate`): replace `pixi install -v` with `pixi lock`.
2. **Environment solving** (`executor.SolveEnvironment` / `JobTypeUpdate`): replace `pixi install -v` with `pixi lock`.

Package syncing (`SyncPackagesFromWorkspace`) continues to call `pixi list`.

Size tracking (`UpdateWorkspaceSize` / `SizeBytes` column) is removed from the workspace table. The `size_bytes` field in API responses becomes optional (`json:"size_bytes,omitempty"`) — clients must adapt. GORM `AutoMigrate` will not drop the column; it becomes an orphan. An explicit `DROP COLUMN` migration is optional future cleanup. In local mode, installed status and on-disk size are derived from disk asynchronously and must never block workspace listing.

### Workspace status (all modes)

`WsStatusReady` is redefined: it now means "locked and ready for install" rather than "installed and ready to use." No new status value is added.

All existing `WsStatusReady` gates are safe with this redefinition:
- **Rollback** (`workspace_worker.go`): only needs to restore a previous `pixi.toml` and `pixi.lock`.
- **OCI push** (`workspace.go`): already operates from `pixi.toml` + `pixi.lock`.
- **OCI publish** (`workspace_publishing.go`): same as push.
- **Package listing** (`workspace_packages.go`): reads from lockfile (verified above).

The presence of `.pixi/envs/` means the workspace is installed. This also handles the upgrade path naturally — existing workspaces with `.pixi/envs/` already on disk will be detected as installed.

#### API endpoints

- **`POST /api/workspaces/:name/install`** — enqueues an install job (type `JobTypeInstall`), returns the job ID immediately. The job runs `pixi install` (which installs packages from the already-present `pixi.lock` into `.pixi/envs/`). Progress is streamed through the existing SSE log pipeline. Does not block.
- **`POST /api/workspaces/:name/uninstall`** — enqueues an uninstall job (type `JobTypeUninstall`), returns the job ID immediately. The job removes `.pixi/envs/` and clears cached environment metadata.

#### Install status in API response

Install status is an inline field on `WorkspaceResponse`, following the same pattern as `WorkspaceStatus` — a string enum with both steady and in-progress states:

| Value | Meaning |
|---|---|
| `not_installed` | Lockfile present, no `.pixi/envs/` |
| `installing` | Install job enqueued or running |
| `installed` | `.pixi/envs/` present on disk |
| `uninstalling` | Uninstall job enqueued or running |
| `install_failed` | Last install job failed |

#### UI

- An "Install" button on each workspace with `install_status: "not_installed"` or `"install_failed"`. The button is hidden or replaced with a progress indicator when `install_status` is `"installing"` or `"uninstalling"`.
- An "Uninstall" button on each workspace with `install_status: "installed"`.
- Each workspace card shows an install status badge.

#### CLI

- `nebi workspace install <name>` — enqueues an install job and streams progress via the existing SSE log pipeline.
- `nebi workspace uninstall <name>` — enqueues an uninstall job.
- `nebi workspace list`:
  - **Default (no flag):** lists all workspaces, indicating install status per workspace (e.g., a `[installed]` tag).
  - **`--installed`:** filters to only workspaces with `install_status: "installed"`.

A workspace that is not yet installed can still be browsed (package list, metadata), published to OCI, and deleted. Only operations that require the actual environment files are gated on install status.

### Workspace updates and rollback (local mode only)

The lockfile update (`pixi lock`) applies to all modes; the auto-install behavior described below is local-mode only.

Any change to the workspace (manifest edits via add/remove packages, or rollback to a previous version) restores or updates `pixi.toml` and triggers `pixi lock` to produce a new `pixi.lock`. A new workspace version snapshot is created.

After the lockfile is updated, install status from before the operation is respected:

- If the workspace was installed before, automatically run `pixi install` to keep the environment in sync with the new lockfile.
- If the workspace was not installed, stop after the lockfile update (no automatic install).

The "was installed before" check uses disk discovery (`.pixi/envs/` exists before the operation begins).

### What does not change

- **OCI publish/pull:** already operates on `pixi.toml` + `pixi.lock` only. No changes.

## Tradeoffs & Risks

| Tradeoff | Mitigation |
|---|---|
| Workspaces default to not-installed; users must take an explicit step. | The install button is prominent in the UI. The CLI `workspace list` output makes status visible. This is the entire point of the change. |
| `WsStatusReady` now conflates two states (locked vs. installed). | Acceptable for MVP — all existing gates are safe with the redefinition (see Workspace status above). A follow-up client refactor will split these states properly. |
| Environment metadata (install status + size) is discovered from disk. Walking every `.pixi/envs/` on every workspace-list call would be slow on cold start. | Size computation must happen asynchronously and must never block workspace listing. If cold-start performance is unacceptable, add a persistent cache (local table or file) updated on install, uninstall, and workspace delete. Cache drift from manual `.pixi/envs/` deletion is a risk of that approach. |

## Open Questions

- **In-progress state transitions and race conditions.** How does `install_status` transition between states — is there a job callback, polling loop, or SSE event that updates the field? How is double-install prevented at the queue level (e.g., rejecting a second install request while a job is already enqueued)? What happens if a user clicks Uninstall while an auto-install from a manifest update is in progress?

- **Error handling for install failures.** What does the user see when `pixi install` fails (network error, corrupt lockfile, disk full)? Does a failed install leave partial `.pixi/envs/` content that could confuse the next attempt? Is there a retry path, or does the user click Install again? The `install_failed` status handles the steady-state case but the transition and cleanup logic needs definition.

aktech:
Server-side create, solve, and rollback now run pixi lock instead of pixi install, so workspaces resolve to a lockfile without downloading packages. Installing the environment is an explicit step: new install/uninstall endpoints and jobs, derived install_status, nebi workspace install/uninstall CLI commands with streamed progress, list --installed, and Install/Uninstall buttons in the UI (local mode only). Updates and rollbacks auto-reinstall when the environment was installed before.

Screenshot 2026-07-16 at 15 09 23

Closes #438


internal/worker/worker.go
```diff
@@ -395,6 +427,23 @@ func (w *Worker) executeJob(ctx context.Context, job *models.Job, logWriter io.W
...
+// operation, but only in local mode and only when the workspace had an
+// installed environment before the operation. This keeps installed
+// environments in sync with the latest lockfile without ever implicitly
+// installing a workspace the user never installed.
+func (w *Worker) maybeReinstallEnv(ctx context.Context, ws *models.Workspace, wasInstalled bool, logWriter io.Writer) error {
+	if !w.svc.IsLocal() || !wasInstalled {
+		return nil
+	}
+	fmt.Fprintf(logWriter, "Workspace was installed; reinstalling environment from updated lockfile...\n")
+	if err := w.executor.InstallEnvironment(ctx, ws, logWriter); err != nil {
```
jbouder, line 446: The manifest, lockfile, and version snapshot are already committed before this runs, so the user's change succeeded. Returning the error marks the whole job failed and leaves the workspace Ready.

Suggested Fix: don't fail the job here. Log it, set install_status = install_failed, and let the job complete.

internal/service/workspace_env.go
```diff
@@ -0,0 +1,76 @@
...
+		Order("created_at DESC").
+		First(&latest).Error
+
+	if jobErr == nil && (latest.Status == models.JobStatusPending || latest.Status == models.JobStatusRunning) {
+		if latest.Type == models.JobTypeEnvInstall {
+			return models.InstallStatusInstalling
+		}
+		return models.InstallStatusUninstalling
+	}
+	if s.executor.IsEnvInstalled(ws) {
```
jbouder: IsEnvInstalled only checks that .pixi/envs exists, but a failed pixi install often creates that dir and then fails partway. Because this runs before the failed-job branch below, a broken install reports installed and install_failed becomes unreachable, so the UI shows "Installed" for a broken env with no retry option.

Fix: move the failed-env_install branch above this check so a failed install wins over bare directory existence.

jbouder (changes requested):
Local testing looks good. A few minor findings in review.

internal/worker/worker.go
aktech: Fixed, now we add the status as install failed, when save succeeds but install fails:

Screenshot 2026-07-23 at 13 25 02

internal/service/workspace_env.go
aktech: Fixed, now we look up the workspace's most recent install/uninstall job from the database first.

jbouder (approved):

PR #458, opened 2026-07-15
security: harden OIDC state cookie flags and drop auth PII from logs

### LINKED ISSUE #454: [security] OIDC state cookie lacks Secure/SameSite; auth logs retain state values and PII
## Summary

The OIDC state cookie is set and cleared with `Secure=false` and no explicit `SameSite`, and the authentication path logs raw state values and personally identifying claims. The random state itself is strong and checked on callback (a good control); this finding is about cookie flags and log hygiene. (CWE-614 Sensitive Cookie in HTTPS Session Without Secure Attribute.)

- State cookie set with `secure=false`: `internal/api/handlers/oidc.go:32` (`c.SetCookie("oidc_state", state, 600, "/", "", false, true)`) and cleared the same way at `:64`. No `SameSite` is configured anywhere (`SetSameSite`/`SameSite` absent from `internal/`).
- Invalid callback logs both the supplied and stored state: `internal/api/handlers/oidc.go:58` (`slog.Warn("Invalid OIDC state", "state", state, "stored_state", storedState)`).
- Login logs email, name, subject, and picture at info level: `internal/auth/oidc.go:143` (`slog.Info("OIDC login claims", "email", ..., "name", ..., "sub", ..., "picture", ...)`).
- Positive control retained: state is 32 random bytes (`generateRandomState`, `oidc.go:102-108`) and compared on callback (`:57`). (The comparison is a plain `!=`, not constant-time - a minor extra nit.)

## Impact

- The state cookie may be sent over cleartext HTTP where such a route is reachable.
- Authentication state and PII are unnecessarily retained in logs.
- The missing explicit SameSite policy reduces defense in depth.

## Remediation

- Set `Secure`, `HttpOnly`, and an explicit appropriate `SameSite` on the state cookie, deriving cookie security from the externally visible HTTPS origin rather than only backend transport.
- Do not log raw state values or unnecessary identity claims; use structured redaction and retention limits.

_Identified in a security assessment of v0.13 (commit `be3187c`), re-validated against `main` @ `15f66d4`. Confidence: High. Does not on its own claim the OIDC flow is CSRF-bypassable._

aktech:
Sets `SameSite=Lax` and `Secure` (derived from the request's HTTPS origin, including `X-Forwarded-Proto`) on the `oidc_state` cookie via one shared helper, and stops logging raw state values and OIDC identity claims.

Closes #454

### DIFF
--- a/internal/api/handlers/oidc.go
+++ b/internal/api/handlers/oidc.go
@@ -6,10 +6,33 @@ import (
 	"encoding/json"
 	"log/slog"
 	"net/http"
+	"strings"
 
 	"github.com/gin-gonic/gin"
 	"github.com/nebari-dev/nebi/internal/auth"
 )
+
+// isRequestHTTPS reports whether the request reached Nebi over HTTPS, as seen
+// by the client. It trusts the terminating proxy's X-Forwarded-Proto header
+// (the common Kubernetes/ingress deployment terminates TLS upstream) and falls
+// back to the direct TLS connection state. Used to set the Secure flag on
+// auth cookies from the externally visible origin, not just backend transport.
+func isRequestHTTPS(c *gin.Context) bool {
+	if c.Request.TLS != nil {
+		return true
+	}
+	return strings.EqualFold(c.GetHeader("X-Forwarded-Proto"), "https")
+}
+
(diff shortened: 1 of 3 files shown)


jbouder (approved):
Checks out

PR #462, opened 2026-07-20
security: reject weak JWT secrets in team mode, derive signing key separately

### LINKED ISSUE #443: [security] Default JWT secret ships and is not rejected at startup; Helm sets no secret
## Summary

Nebi ships a public default value for the JWT signing secret and never rejects it at startup, and that same secret is the root input used to derive the encryption key for stored registry credentials. A standard Helm install does not set the secret, so a team deployment inherits the public default. (CWE-798 Use of Hard-coded Credentials; related CWE-321 Use of Hard-coded Cryptographic Key.)

- Config defaults `auth.jwt_secret` (env `NEBI_AUTH_JWT_SECRET`) to the literal `change-me-in-production`: `internal/config/config.go:99`.
- The basic authenticator signs 24-hour JWTs with HS256 using that secret; the middleware trusts the token's user UUID after loading that user from the database: `internal/auth/basic.go` (`TokenDuration = 24 * time.Hour` at `:26`; `jwt.SigningMethodHS256` at `:122`; UUID load at `:223-229`).
- The credential-encryption key is derived from the **same** secret: `internal/api/router.go:178` `nebicrypto.DeriveKey(cfg.Auth.JWTSecret)` (and the worker path `internal/server/server.go:158`). The KDF is HKDF-SHA256 with domain separation (`internal/crypto/encrypt.go:16-38`), so the derived key is not literally the signing key - but the **root secret input is the public default**, so knowledge of it yields both.
- No startup validation rejects an empty/default/weak secret in team mode. The only guard is an empty-string check inside `DeriveKey` (`internal/crypto/encrypt.go:28`), which the non-empty default never trips.
- The chart never sets the secret: neither `chart/values.yaml`, `chart/values-prod.yaml`, `chart/templates/deployment-api.yaml`, nor `chart/templates/deployment-worker.yaml` exports `NEBI_AUTH_JWT_SECRET`, and `chart/templates/secrets.yaml` holds only the Postgres password.

## Impact

- A party who learns the default secret can forge a valid 24-hour token for any user whose UUID they know.
- Encrypted registry credentials become vulnerable to offline decryption if the database is disclosed, because the key-derivation input is publicly known.
- Because signing and credential encryption share the same root secret, rotating the JWT secret also changes the encryption key, so existing encrypted registry credentials require explicit migration or re-entry.

## Remediation

- Fail startup in team mode if the secret is empty, the known default, or weak.
- Add a Helm Secret for a randomly generated JWT key, or require an externally managed Secret; wire it into the API and worker deployments.
- Use separate, independently generated keys for JWT signing and for registry-credential encryption.
- Support key identifiers and overlapping rotation windows; implement versioned registry-ciphertext migration.

## Acceptance criteria

- A default chart render contains no public cryptographic secret, and install fails clearly when no secure secret source is configured.
- JWT signing and registry encryption use different key material.
- Tokens signed with `change-me-in-production` are rejected.
- Rotation preserves registry access while invalidating old JWTs.

aktech:
Team-mode startup now fails if `auth.jwt_secret` is empty, the shipped default (`change-me-in-production`), or under 32 characters, so a deployment can no longer silently inherit a public secret. The JWT signing key is derived from that secret via HKDF with its own domain-separation string, independent from the key used to encrypt stored registry credentials, so a token forged with the raw secret no longer validates.

Local mode is unaffected: it never reaches the JWT auth path.

Closes #443


jbouder (approved):
Only thing to mention is that when deployed, this change will invalidate all existing JWTs, so everyone will need to log in again. Super low risk though.

PR #463, opened 2026-07-21
Enforce registry RBAC on publish and registry access paths

### LINKED ISSUE #444: [security] Registry read/write RBAC is defined but never enforced on browse/tags/import/publish
## Summary

The RBAC layer defines registry read/write permissions and lets administrators grant them to groups, but those permissions are never checked on the user-facing registry routes. Browse, tags, import, and publish all use Nebi's stored registry credentials without verifying that the caller holds the corresponding registry permission. (CWE-862 Missing Authorization.)

- The permissions exist and are grantable: `CanReadRegistry` / `CanWriteRegistry` at `internal/rbac/rbac.go:274,282`, exposed on the provider at `internal/rbac/provider.go:10-11`; admin grant routes at `internal/api/router.go:311-312`.
- A repo-wide search finds **zero** production call sites of `CanReadRegistry` / `CanWriteRegistry` - they are referenced only in tests (`internal/service/admin_test.go`, `internal/rbac/rbac_test.go`).
- The following routes carry only group-level `authenticator.Middleware()` and no registry authorization:
  - `GET /registries/:id/repositories` → `registry_browse.go` `ListRepositories` (`router.go:262`)
  - `GET /registries/:id/tags` → `ListTags` (`router.go:263`)
  - `POST /registries/:id/import` → `ImportEnvironment` (`router.go:264`)
  - `POST /workspaces/:id/publish` → guarded by `RequireWorkspaceAccess("write")` only, no registry check (`router.go:241`)
- The browse handler loads **decrypted** registry credentials and uses them for catalog/tag requests with no read check: `internal/api/handlers/registry_browse.go:41,53` (and `:130,141`); decryption at `internal/service/registry.go:237-242`.
- Publish and import load the registry endpoint and use its credentials without a registry write/read check: `internal/service/workspace_publishing.go:39-86`, `internal/service/workspace_import.go:50-73`.

## Impact

- Disclosure of private repository names, tags, and environment content to users with no registry grant.
- Pulling private registry artifacts outside intended group boundaries.
- Pushing / publishing to a registry using Nebi's shared service credential without registry write permission.
- Audit records that show an operation occurred but not that the required registry authorization was present.

## Remediation

- Require registry **read** permission for browse, tags, and import; require registry **write** permission for publish and visibility changes.
- Enforce the check inside the service method (defense in depth), not only in middleware, and pass the authenticated subject into every registry domain operation.
- Ensure unauthorized users cannot enumerate private registry IDs or metadata; audit denied registry access.

## Acceptance criteria

- A user with no registry grant receives `403` for browse, tags, import, publish, and visibility changes.
- Read permission does not permit publishing.
- Group removal takes effect immediately.
- Tests cover direct service invocation as well as HTTP routes.

MUFFANUJ:
## Reference Issues or PRs

Closes #444 

- [x] Documentation Update
- [ ] Build related changes

## Testing

- [x] Did you add new tests?

## Documentation
API documentation was regenerated to include the new `registry_id` query parameter and the documented `400`/`403` responses for publish defaults

#### Text styling

#### Non-text content

## Any other comments?

This fixes missing registry RBAC enforcement across registry browse, tag lookup, import, publish, publication visibility updates, and publish defaults. Registry read/write checks are now performed in the service layer, denied access is audited, registry listings/publications are filtered by readable registries, and publish defaults can be computed for an explicitly selected registry. The frontend publish dialog now refreshes defaults when the user selects a different registry, and the API docs were regenerated for the new publish-defaults behavior.


MUFFANUJ:
Currently, I kept "read" / "write" as string literals because that is the existing pattern used across the repo’s RBAC code. I think that constants would be cleaner and safer long-term. We can add shared constants for these actions and replace usages repo-wide, but I avoided doing that in this fix to keep the change minimal and focused on registry RBAC enforcement. If needed i can include that as well here.

aktech:
Looks like there are some conflicts.

MUFFANUJ:
Resolved the conflicts in [91e5e72](https://github.com/nebari-dev/nebi/pull/463/commits/91e5e7271da1c77d1c8d6b9f271b18648c80936c). Please take a look now!

aktech:
Looks like the CI is failing after the latest merge.

internal/service/workspace_publishing.go
```diff
@@ -200,12 +217,19 @@ func (s *WorkspaceService) GetPublishDefaults(wsID string) (*PublishDefaultsResu
...
+	}
+	if err := registryQuery.First(&registry).Error; err != nil {
 		if err == gorm.ErrRecordNotFound {
 			return nil, ErrNotFound
 		}
 		return nil, err
 	}
+	if err := ensureRegistryAccess(s.db, s.rbac, s.isLocal, userID, registry.ID, "read"); err != nil {
+		return nil, err
+	}
```
aktech, line 238: We are returning 404 when the registry doesn't exists (correct behaviour) and 403 when the user don't have access to it, which means a user who don't have access to a registry can infer if the registry exists in the system or not.

We should not allow the user to infer the state of an object in the system if they don't have access to it, which means a 404 should be returned instead of 403.

MUFFANUJ: That make sense!

aktech (changes requested):

aktech (changes requested):
@MUFFANUJ Taking a closer look, this breaks registry access for everyone in team mode. Previously any logged-in user could pull and push from the registries configured by admins. With this PR, nobody (including admins) has access after upgrade until grants are created, and the only way to create them is the admin grant API by hand: there's no UI for it.

Fine-grained access control was parked for later mainly because the management side/UI wasn't there yet. How are you envisioning people will grant permissions? We probably need either a backwards-compatible default (e.g. registries stay open to logged-in users unless marked restricted) or the grant UI as part of this change.

I would suggest we go with ability keeping registries open to logged in users unless marked restricted to unblock this PR and a follow up PR for the management UI, for which we need to coordinate with Smera and Nat.


Maybe something like:

---
Screenshot 2026-08-11 at 13 16 41

MUFFANUJ:
> I would suggest we go with ability keeping registries open to logged in users unless marked restricted to unblock this PR and a follow up PR for the management UI, for which we need to coordinate with Smera and Nat.

That makes sense, will address the UI part on a follw up; this is ready for another look!

aktech:
There are some conflicts now.

MUFFANUJ:
@aktech, ready for a look!

aktech (approved):

PR #467, opened 2026-07-28
Add federated identity review flow for safe OIDC account linking

### LINKED ISSUE #446: [security] OIDC/proxy accounts linked by mutable username/email instead of (issuer, subject)
## Summary

After validating an OIDC (or proxy) token, Nebi links it to a local user record by mutable `username`/`email` claims rather than the immutable `(issuer, subject)` pair. The immutable subject is never persisted or used as the lookup key, and `email_verified` is parsed but never enforced. Where an identity provider allows a colliding or changeable username/email, this permits linking a token to an existing account. (CWE-287 Improper Authentication.)

- OIDC user lookup is `username = ? OR email = ?`: `internal/auth/oidc.go:182`. The `username` is derived from mutable claims (`oidc.go:146-152`).
- The OIDC `sub` is passed to `findOrCreateUser` but never referenced in the body, never persisted, and not part of the lookup or the created record: `internal/auth/oidc.go:178,197-204`.
- `EmailVerified` is declared (`oidc.go:132`) but never read anywhere.
- The user model has no issuer/subject columns: `internal/models/user.go:11-20`.
- The proxy path behaves identically - lookup by preferred username / email, no issuer+subject binding: `internal/auth/proxy.go:61-79`.

## Impact

A token for an attacker-controlled IdP identity can be linked to an existing Nebi user record, inheriting that record's workspace ownership/collaboration, group-derived registry permissions, publication history, and administrator role where present. This is an account-takeover class failure whenever a claim collision is possible.

## Exploit prerequisites

At least one of: the IdP permits selecting/changing an unverified email; preferred usernames are mutable or reusable; a local/basic account collides with an OIDC username/email; issuers/realms are combined without a controlled migration.

## Remediation

- Add a federated-identity table keyed by `(issuer, subject)` and treat that pair as the only automatic lookup key.
- Store email/username/name/picture as mutable profile attributes only.
- Require verified email where email is used; never silently link an OIDC identity to a local account by email or username - require confirmation from both identities or admin approval.
- Provide a migration that detects existing collisions rather than silently selecting a record.

## Acceptance criteria

- Changing email or preferred username does not change the bound account.
- Two issuers with the same subject or email remain distinct.
- An unverified email cannot claim an existing local account.
- Tests cover issuer changes, subject changes, recycled usernames, duplicate email, and local/OIDC collisions.

_Identified in a security assessment of v0.13 (commit `be3187c`), re-validated against `main` @ `15f66d4`. Confidence: High for the implementation flaw; exploitability depends on IdP claim policy._

MUFFANUJ:
## Reference Issues or PRs
Closes #446 

- [x] Documentation Update
- [ ] Build related changes

## Testing

- [x] Did you add new tests?

#### Text styling

#### Non-text content

## Any other comments?

Implements safe federated identity handling for OIDC/proxy/device auth.

- Adds `federated_identities` binding by immutable `issuer + subject`.
- Adds `federated_identity_reviews` for username/email collisions instead of auto-linking existing users.
- Blocks repeat login for rejected identity-link requests.
- Keeps provider email on `users.email` when present, even if unverified; we could add a federated domain here if needed, but it would make it more complex.
- Falls back to `username@nebi.local` or `subject@nebi.local` only when provider email is missing.
- Preserves `email_verified` on federated identity records for admin/UI review.
- Adds admin APIs to list, approve, and reject federated identity reviews.
- Adds remote admin proxy/client support for identity reviews.
- Adds admin UI page for reviewing, approving, and rejecting identity requests.
- Shows pending/rejected login messages for blocked federated login attempts.
- Adds dashboard/audit log integration for identity review activity.
- Adds DB migration coverage and legacy federated-user guard.
- Regenerates Swagger docs.
- Adds backend and frontend tests for collision review, approval, rejection, rejected retry, and login messages.

Basically, it implements the security fix with the approval review workflow; UI verification was done locally with SQLite by running Nebi in team mode against a temporary SQLite DB, seeding local alice/bob users, and using a mock OIDC provider to simulate federated logins. See below:

https://github.com/user-attachments/assets/3f09f74a-8874-4bfc-b966-9d82493a7c7b

Logs:

Screenshot 2026-07-28 at 1 56 05 PM


internal/db/db.go
```diff
@@ -131,6 +137,25 @@ func Migrate(db *gorm.DB) error {
 	return nil
 }
 
+func detectLegacyFederatedUsers(db *gorm.DB) error {
```
tylerpotts: **Blocker:** This makes `Migrate` return an error whenever any user has an empty `password_hash` and no `federated_identities` row - i.e. every pre-existing team-mode deployment that ever used OIDC or proxy auth will refuse to boot after upgrading, and the operator gets nothing but raw user UUIDs in the error.

Failing closed is a reasonable choice for a security fix, but it needs a remediation path in the same PR. Options, in descending order of preference:

1. A backfill command (e.g. `nebi admin link-federated-identity --user <id> --issuer <iss> --subject <sub>`) referenced from the error message.
2. Auto-creating pending `federated_identity_reviews` rows for legacy users so admins can resolve them through the new UI instead of at the DB console.
3. At minimum, a documented manual procedure (the exact `INSERT INTO federated_identities ...` an operator should run) linked from the error text.

Two related notes: the PR checklist marks this as non-breaking, which this isn't; and the detection heuristic relies on local mode's `local-user` using `"-"` rather than `""` as its hash sentinel (`internal/auth/local.go`) - that cross-package invariant is now load-bearing and security-relevant, so it deserves a comment at both sites.

frontend/src/pages/Login.tsx
```diff
@@ -46,12 +69,29 @@ export const Login = ({ isDarkMode }: LoginProps) => {
     }
     const logoutUrl = useModeStore.getState().logoutUrl;
     if (logoutUrl) {
-      window.location.href = `${getBasePath()}/auth/session`;
+      const checkProxySession = async () => {
+        try {
+          const { data } = await apiClient.get('/auth/session');
```
tylerpotts: **Blocker:** This replaces the documented gateway auto-login flow (browser redirect to `${basePath}/auth/session` -> `SessionRedirect` -> single-use code -> `/auth/code/exchange`) with an XHR to `/api/v1/auth/session`. The comment this change deletes explicitly said `/auth/session` lives *outside* `/api/` "so gateway preserves cookies" - the redirect flow exists because some gateway topologies only attach the IdToken cookie on non-`/api/` paths.

This is an auth-architecture change beyond the scope of #446, and the pending/rejected messaging it enables is already achievable with the redirect flow: this PR teaches `SessionRedirect` to redirect to `/login?error=identity_review_pending|rejected`, which this component already handles via `searchParams`.

Can this be reverted to the redirect flow, or is there a concrete reason the XHR is needed? If it stays, it needs verification against the Envoy/Keycloak gateway topology the original comment was written for.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,312 @@
...
+	claims.Issuer = strings.TrimSpace(claims.Issuer)
+	claims.Subject = strings.TrimSpace(claims.Subject)
+	claims.PreferredUsername = strings.TrimSpace(claims.PreferredUsername)
+	claims.Email = strings.TrimSpace(claims.Email)
+	claims.Name = strings.TrimSpace(claims.Name)
+	claims.AvatarURL = strings.TrimSpace(claims.AvatarURL)
+	return claims
+}
+
+func federatedReviewCollision(db *gorm.DB, claims federatedUserClaims) (*models.User, string, error) {
```
tylerpotts: **Non-blocking (follow-up ok):** Collision detection uses exact `username = ?` / `email = ?` matches, and `normalizeFederatedClaims` only trims whitespace. An IdP claim of `Alice` or `ALICE@example.com` won't collide with a local `alice` / `alice@example.com`, so it skips the review gate and creates a separate confusable account.

Not an account takeover - the (issuer, subject) binding holds - but it partially bypasses the review gate's intent. Suggest case-insensitive comparison (and probably lowercasing in `normalizeFederatedClaims`) for collision detection. Happy for this to be a follow-up issue if you'd rather keep this PR focused.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	return nil, "", nil
+}
+
+func recordFederatedIdentityReview(db *gorm.DB, userID uuid.UUID, field string, claims federatedUserClaims) error {
+	var review models.FederatedIdentityReview
+	err := db.Where("issuer = ? AND subject = ?", claims.Issuer, claims.Subject).First(&review).Error
+	if err == nil {
+		if review.Status == models.FederatedIdentityReviewStatusRejected {
```
tylerpotts, line 260: **Non-blocking question:** Rejection is only enforced here, inside the collision path - so a rejected identity is blocked only while the colliding local user still exists with the colliding claim. If that user is later renamed or deleted, the previously rejected identity silently gets a fresh account on its next login.

That may be acceptable (once the collision is gone there's no takeover risk), but admins might reasonably expect "rejected" to mean blocked outright. Worth either checking for a rejected review before the collision branch, or documenting the intended semantics.

internal/models/federated_identity.go
```diff
@@ -0,0 +1,68 @@
...
+	FederatedIdentityReviewStatusRejected = "rejected"
+)
+
+// FederatedIdentity binds an external OIDC identity to a Nebi user.
+type FederatedIdentity struct {
+	ID            uuid.UUID      `gorm:"type:text;primary_key" json:"id"`
+	UserID        uuid.UUID      `gorm:"type:text;not null;index" json:"user_id"`
+	User          User           `gorm:"foreignKey:UserID;constraint:OnDelete:CASCADE" json:"-"`
+	Issuer        string         `gorm:"not null;uniqueIndex:idx_federated_identity_issuer_subject" json:"issuer"`
+	Subject       string         `gorm:"not null;uniqueIndex:idx_federated_identity_issuer_subject" json:"subject"`
```
tylerpotts: Nice - the composite unique index on `(issuer, subject)` means even a race between two concurrent first logins can't produce a duplicate binding, and the approve flow re-checks for an existing identity inside its transaction before linking. Defense in depth done right.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,312 @@
...
+	Issuer            string
+	Subject           string
+	PreferredUsername string
+	Email             string
+	EmailVerified     bool
+	Name              string
+	AvatarURL         string
+}
+
+func findOrCreateFederatedUser(db *gorm.DB, claims federatedUserClaims) (*models.User, error) {
```
tylerpotts, line 49: This single entry point for all three auth paths (OIDC callback, proxy cookie/session, device flow) is the structural fix that makes the whole thing hold together - the original vulnerability existed partly because the paths had diverged. The device-flow path previously never set issuer/subject at all, so pulling it through here closes a gap beyond what #446 strictly asked for.

tylerpotts (changes requested):
Thanks for this - the security design here is genuinely good. The `(issuer, subject)` binding, the fail-closed review workflow (review row created before any user exists, so a rejected identity never gets an account), and test coverage that maps one-to-one onto the acceptance criteria in #446 all close the vulnerability properly. I verified the full backend suite and frontend suite pass locally, and all three auth paths (OIDC callback, proxy cookie/session, device flow) now route through the shared provisioning function.

I'm requesting changes on two items that are about operational packaging rather than the security design itself:

1. **The migration hard-fails at boot for every existing deployment with OIDC/proxy users, with no remediation path** (inline comment on `internal/db/db.go`). Failing closed is defensible for a security fix, but it needs a backfill command or at minimum a documented manual procedure shipped in the same PR - otherwise this converts a latent vulnerability into a guaranteed outage for exactly the deployments it protects. The PR is also marked "non-breaking", which this isn't.

2. **The gateway auto-login flow is rewritten from redirect+code to XHR** (inline comment on `frontend/src/pages/Login.tsx`). This is an auth-architecture change beyond the issue's scope, and the deleted comment explains why the old design existed. It should be reverted or explicitly justified for the Envoy/Keycloak topology.

Everything else is non-blocking - a couple of follow-up-worthy questions inline, and some stylistic notes in a separate comment. Fix the migration story and resolve the login-flow question and this is an easy approve.

tylerpotts:
A few general notes to go with the review - none of these block approval:

**Architecture:** The change fits the codebase's layering well: handler -> service -> models for the admin flow, auth-package provisioning matching the pre-existing pattern, transactional audit logging, typed service errors, regenerated swagger, cliclient mirroring, and the frontend api/hook/page structure all follow existing conventions. Supply-chain-wise it's clean too: no new dependencies, and the file changes all match the stated scope.

**Duplication worth consolidating (follow-up ok):**
- The pending-status predicate (`status == "" || status == "pending"`) is re-implemented in `internal/service/admin.go`, `FederatedIdentityReviews.tsx`, and `AdminDashboard.tsx`. A method on `models.FederatedIdentityReview` would give the pending/rejected state machine one owner.
- The pending/rejected error-to-response mapping is copy-pasted six times across `internal/api/handlers/auth.go` (x3), `deviceflow.go`, `oidc.go`, and `internal/auth/basic.go`. A small shared helper (JSON variant + redirect variant) would keep the next auth error state from requiring six edits.
- `Login.tsx` matches on the backend's literal error message strings as a fallback alongside the `identity_review_*` codes. Matching on message text is brittle - prefer emitting the machine code from the JSON response paths too.

**Small observations:**
- Approve logs audit resource `federated_identity:<id>` while reject logs `federated_identity_review:<id>` - defensible since they act on different entities, but worth a comment since audit consumers filter by resource prefix.
- The empty-email fallback domain changed from `@proxy.local` to `@nebi.local`, so pre-existing users keep the old domain while new ones get the new one. Harmless, but flagging in case it wasn't intentional.

Verification: full backend suite (`go test ./internal/...`) and frontend suite (198/198 via vitest) both pass on `ad85b45`, and I traced the acceptance criteria from #446 to specific tests - they're all covered, several more strictly than the issue asked for.

internal/models/federated_identity.go
MUFFANUJ: Thanks!

internal/auth/federated_identity.go
MUFFANUJ: Updated this so rejected reviews are checked by stable `(issuer, subject)` before any `username`/`email` collision logic. That means a rejected external identity stays blocked even if the original local user is later `renamed`/`deleted` or the mutable claims change

internal/auth/federated_identity.go
MUFFANUJ: Added case-insensitive `username`/`email` collision checks and normalized incoming mutable claims to lowercase, so `Alice` / `ALICE@example.com` now collides with existing `alice` / `alice@example.com` and goes through the review gate instead of creating a confusable account.

internal/db/db.go
MUFFANUJ: Added a remediation path in the latest commit.

frontend/src/pages/Login.tsx
MUFFANUJ: Reverted.

MUFFANUJ:
Thanks for the detailed review! @tylerpotts, I have addressed the actionable cleanup here; it's ready for another review

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,351 @@
...
```
tylerpotts, line 49: This single entry point is what makes the fix hold. I grepped the tree: `username = ? OR email = ?` is gone, and the only remaining user-creation sites are this function and admin `CreateUser`. OIDC callback, proxy cookie, and device flow all land here, which is what the original issue asked for.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,351 @@
...
+
+func recordFederatedIdentityReview(db *gorm.DB, userID uuid.UUID, field string, claims federatedUserClaims) error {
+	var review models.FederatedIdentityReview
+	err := db.Where("issuer = ? AND subject = ?", claims.Issuer, claims.Subject).First(&review).Error
+	if err == nil {
+		if review.Status == models.FederatedIdentityReviewStatusRejected {
+			return fmt.Errorf("%w: issuer %s subject %s", errFederatedIdentityRejected, claims.Issuer, claims.Subject)
+		}
+		updates := map[string]any{
+			"user_id":         userID,
```
tylerpotts: **Blocker:** this overwrites `user_id` on an existing pending review, keyed only on `(issuer, subject)`. Combined with `ApproveFederatedIdentityReview` reading `review.UserID` fresh from the DB and taking only a review ID from the client, the admin's approval isn't bound to the claims they actually reviewed.

Concretely: log in once with `preferred_username: "alice"` to file an innocuous review. Admin opens `/admin/identity-reviews`. Before they click Approve, log in again with the same `(issuer, subject)` but `preferred_username: "root-admin"`. Same row, `user_id` now points at the admin account, status still pending, `reviewed_by` reset to nil. The admin approves the row they were looking at and binds the external identity to the admin account.

I confirmed this against this commit with a throwaway test:

```
admin sees review b8eec015... -> user 7996fe63... (alice)
review rows total: 1
SWAPPED: review b8eec015... now targets admin 9e62313b... (root-admin)
```

Simplest fix: don't mutate a pending review at all. If a row already exists for this `(issuer, subject)` and is pending, leave it and return the pending error. The admin then approves exactly what was filed, and nothing is lost because profile fields refresh on the next successful login via `updateFederatedIdentityProfile`. Alternatively, add optimistic concurrency: have the approve request carry the expected `user_id` (or `updated_at`) and return 409 on mismatch.

internal/service/admin.go
```diff
@@ -302,6 +303,107 @@ func (s *AdminService) ListAuditLogs(userIDFilter, actionFilter string) ([]model
...
+		err := tx.Where("issuer = ? AND subject = ?", review.Issuer, review.Subject).First(&existing).Error
+		if err == nil {
+			return &ConflictError{Message: "federated identity is already approved"}
+		}
+		if !errors.Is(err, gorm.ErrRecordNotFound) {
+			return err
+		}
+
+		identity = models.FederatedIdentity{
+			UserID:        review.UserID,
```
tylerpotts, line 406: Companion to the `recordFederatedIdentityReview` comment: `review.UserID` is read fresh inside the transaction and the endpoint takes only a review ID, so there's nothing tying this write to what the admin saw in the UI. Whichever fix you pick for the other comment, this is the line that consumes the swapped value.

internal/service/admin.go
```diff
@@ -302,6 +303,107 @@ func (s *AdminService) ListAuditLogs(userIDFilter, actionFilter string) ([]model
...
+	var identity models.FederatedIdentity
+	if err := s.db.Transaction(func(tx *gorm.DB) error {
+		var review models.FederatedIdentityReview
+		if err := tx.Preload("User").First(&review, "id = ?", reviewID).Error; err != nil {
+			if errors.Is(err, gorm.ErrRecordNotFound) {
+				return ErrNotFound
+			}
+			return err
+		}
+		if review.User.ID != review.UserID {
```
tylerpotts, line 384: Good guard. A review whose target user was deleted after the review was filed can't be approved into a dangling binding, because the `Preload` returns a zero `User` and this catches it.

internal/db/db.go
```diff
@@ -137,6 +144,30 @@ func Migrate(db *gorm.DB, seedRegistry bool) error {
 	return nil
 }
 
+const legacyFederatedIdentityMigrationDocs = "https://nebi.nebari.dev/docs/server-setup#legacy-federated-identity-migration"
+
+func detectLegacyFederatedUsers(db *gorm.DB) error {
```
tylerpotts: **Blocker (carried over):** this still hard-fails startup for every team-mode deployment that has ever used OIDC, proxy auth, or device flow. The docs help an operator recover, but they ask for hand-written `INSERT`s into a GORM-managed table, with operator-generated UUIDs, in two SQL dialects, against a production DB, while the service is down.

That work is unnecessary, because this PR already ships a better migration. I verified it in a worktree with a throwaway test: a legacy user (username == email == `claims.Email`, empty password hash, no `federated_identities` row) logging in after the upgrade collides with themselves on `users.email`, files a pending review pointing at their own record, and after approval login returns the original account:

```
legacy user resolvable via review flow: review=1c35de1f... field=email
post-approval login returns the original account 3b2bc9c7... (alice@example.com)
```

No downtime, no guessing at issuer/subject values, and the linkage gets audit-logged instead of applied by hand.

Suggestion: drop this from `Migrate` and let the review queue be the migration path, with a note in `server-setup.md` so admins expect a burst of reviews after upgrading. If you want a safety net, log a warning with the count and stay up.

Two design problems go away with it, which is worth noting either way:

* `Migrate` is otherwise only `AutoMigrate` plus idempotent seeders. No other step inspects pre-existing rows and aborts boot, and this one runs on every start of both entry points, including the desktop app that forces `NEBI_MODE=local`, where federated identity can't exist.
* "This user is federated" is inferred from `users.password_hash = ''`, a sentinel owned by `internal/auth`. That's why `internal/auth/local.go:30` now carries a comment whose only job is to satisfy `internal/db`. Meanwhile `federated_identity.go:132` still creates *new* federated users with an empty hash, so the empty-hash class permanently means two things, and any future passwordless-but-not-federated user (invite flow, service account) silently becomes a boot-blocking condition. If you keep the guard, key it on an explicit `provider`/`auth_source` column instead.

internal/db/db.go
```diff
@@ -137,6 +144,30 @@ func Migrate(db *gorm.DB, seedRegistry bool) error {
...
+	}
+	if len(users) == 0 {
+		return nil
+	}
+
+	details := make([]string, len(users))
+	for i, user := range users {
+		details[i] = fmt.Sprintf("id=%s username=%q email=%q", user.ID, user.Username, user.Email)
+	}
+	return fmt.Errorf("legacy federated users without issuer/subject bindings block startup; backfill issuer/subject bindings before upgrading and see %s; affected users: %s", legacyFederatedIdentityMigrationDocs, strings.Join(details, "; "))
```
tylerpotts: If the guard stays: this interpolates id, username, and email for every affected user into a single error string. On a deployment with a few hundred federated users that's an unreadable startup failure carrying PII into log aggregators. It also cuts against the convention two files over at `internal/auth/oidc.go:145` ("Avoid logging claims... they end up in aggregated log stores. A presence marker is enough for ops."). A count plus the first few IDs would serve ops just as well.

internal/service/admin.go
```diff
@@ -302,6 +303,107 @@ func (s *AdminService) ListAuditLogs(userIDFilter, actionFilter string) ([]model
...
+				return ErrNotFound
+			}
+			return err
+		}
+		if !review.IsPending() {
+			return &ConflictError{Message: "federated identity review was already rejected"}
+		}
+
+		now := time.Now().UTC()
+		review.Status = models.FederatedIdentityReviewStatusRejected
```
tylerpotts: Non-blocking: rejection is permanent with no in-product undo. There's no un-reject or delete endpoint, `ApproveFederatedIdentityReview` returns a conflict for a rejected row, and the UI renders rejected rows with no actions. A mis-click in the confirm dialog locks that external identity out forever, recoverable only by SQL. The dialog does warn, so I'm fine with a follow-up issue rather than blocking, but combined with the migration guard the recurring theme is "recovery requires DB surgery."

internal/service/admin.go
```diff
@@ -302,6 +303,107 @@ func (s *AdminService) ListAuditLogs(userIDFilter, actionFilter string) ([]model
...
+			Username:      review.Username,
+			Email:         review.Email,
+			EmailVerified: review.EmailVerified,
+			Name:          review.Name,
+			AvatarURL:     review.AvatarURL,
+		}
+		if err := tx.Create(&identity).Error; err != nil {
+			return fmt.Errorf("approve federated identity: %w", err)
+		}
+		if err := tx.Delete(&review).Error; err != nil {
```
tylerpotts: Non-blocking: approve soft-deletes the review, but the unique index on `(issuer, subject)` has no `deleted_at` component, so the row keeps occupying that index while being invisible to `recordFederatedIdentityReview`'s scoped `First()`. Any future path needing a second review row for the same pair would hit a constraint violation instead of creating one. Unreachable today, since an approved identity short-circuits earlier, but `Unscoped().Delete()` would close it.

Related question on the lifecycle: rejection is a status transition, approval is a row deletion, and there's no `StatusApproved` constant. Was an explicit `approved` status considered, so the review table is the single record of the decision? The UI currently derives a strict Pending/Rejected binary from `IsPending()` and so can never show an approval.

internal/service/admin.go
```diff
@@ -302,6 +303,107 @@ func (s *AdminService) ListAuditLogs(userIDFilter, actionFilter string) ([]model
 	return logs, nil
 }
 
+// ListFederatedIdentityReviews returns federated identity review decisions.
+func (s *AdminService) ListFederatedIdentityReviews() ([]models.FederatedIdentityReview, error) {
```
tylerpotts: No `Limit` here, while the sibling `ListAuditLogs` caps at 100 just above. `AdminDashboard.tsx` fetches the full list on every dashboard load purely to count pending ones, and rejected rows are never deleted, so this only grows.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,351 @@
...
+	return nil
+}
+
+func userByField(db *gorm.DB, field, value string) (*models.User, error) {
+	if field != "username" && field != "email" {
+		return nil, fmt.Errorf("unsupported user collision field: %s", field)
+	}
+
+	var user models.User
+	err := db.Where("LOWER("+field+") = LOWER(?)", value).First(&user).Error
```
tylerpotts, line 326: `LOWER(col) = LOWER(?)` defeats the unique indexes on `users.username` and `users.email`, so every federated login that reaches the create path does two sequential scans. Fine at current scale; worth either storing normalized values or adding a functional index on Postgres if user counts grow.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,351 @@
...
+	if claims.PreferredUsername != "" {
+		return claims.PreferredUsername, nil
+	}
+
+	base := claims.Subject
+	if claims.EmailVerified && claims.Email != "" {
+		base = claims.Email
+	}
+	if base == "" {
+		base = "user-" + suffix
```
tylerpotts: Dead branch: `claims.Subject` is guaranteed non-empty by the check at line 71, so `base` can never be empty here.

internal/auth/basic.go
```diff
@@ -200,6 +200,16 @@ func (a *BasicAuthenticator) Middleware() gin.HandlerFunc {
 		user, err := findOrCreateProxyUser(a.db, proxyClaims)
 		if err != nil {
 			slog.Error("Failed to find/create proxy user", "error", err)
+			if IsFederatedIdentityReviewRequired(err) {
```
tylerpotts: These two branches are the same mapping that `handlers/auth.go` does through the single `FederatedIdentityReviewErrorCode` helper. Two implementations of one mapping introduced in the same PR; this one could call the helper.

internal/audit/audit.go
```diff
@@ -60,6 +60,11 @@ const (
 	ActionLoginFailed       = "login_failed"
 )
 
+const (
```
tylerpotts: Two small consistency points: these could be appended to the existing action block above rather than a new one (the "Audit actions constants" comment now describes only the first block), and no `Resource*` constant was added alongside `ResourceUser` / `ResourceWorkspace` / `ResourcePermission`. That second one matters a little more, because approve logs `federated_identity:<id>` while reject logs `federated_identity_review:<id>`, so the same workflow shows up under two resource types.

frontend/src/pages/Login.tsx
```diff
@@ -13,10 +13,35 @@ type LoginProps = {
   isDarkMode: boolean;
 };
 
+const pendingReviewError = 'federated identity link is pending admin approval';
```
tylerpotts: These two prose branches can never fire: the backend only emits the codes (`identity_review_pending` / `identity_review_rejected`) on every path. The matching exported constants in `federated_identity.go:19-25` are also referenced nowhere in the repo. Worth dropping both so there's one error vocabulary crossing the API boundary rather than two.

internal/auth/proxy_test.go
```diff
@@ -148,6 +233,181 @@ func TestFindOrCreateProxyUser_NoIdentity(t *testing.T) {
...
+	}
+	if review.UserID != localUser.ID {
+		t.Errorf("expected review for local user %s, got %s", localUser.ID, review.UserID)
+	}
+	if review.CollisionField != "email" {
+		t.Errorf("expected email collision field, got %s", review.CollisionField)
+	}
+}
+
+func TestFindOrCreateProxyUser_DistinguishesIssuersWithSameSubject(t *testing.T) {
```
tylerpotts, line 531: The test coverage here maps onto the issue's acceptance criteria one for one: issuer change, subject change, recycled usernames, duplicate email, local/OIDC collision, unverified email. `TestFindOrCreateProxyUser_FindsExisting` in particular asserts that changed username and email claims do not move an existing binding, which is the property the whole change exists to guarantee.

tylerpotts (changes requested):
Thanks for the follow-up round. All three items I raised last time are properly fixed, and I verified them in a worktree: the gateway auto-login redirect is restored, collision detection is case-insensitive on both sides now, and rejection is enforced by `(issuer, subject)` before any collision logic, with a test for the no-current-collision case.

I re-ran everything on this commit: `go build ./internal/... ./cmd/...` is clean, the Go tests pass, `npx vitest run` gives 31 files / 202 tests passing, and `swag init` produces no diff, so the swagger drift check will pass.

The core of #446 is solved. Every external auth path funnels through one function, `username = ? OR email = ?` is gone from the codebase entirely, and the test coverage maps onto the issue's acceptance criteria one for one. That is the hard part and it's right.

Two things are blocking, one new and one carried over.

**New:** there's a confused-deputy hole in the approval flow itself. A pending review can be re-pointed at a different local user in between the moment an admin reads it and the moment they click Approve, so the admin's decision isn't bound to the claims they reviewed. I confirmed it with a throwaway test. Details inline on `recordFederatedIdentityReview`.

**Carried over:** the legacy-user startup guard. I'd argue for dropping it rather than documenting it. I verified that a legacy OIDC user logging in after the upgrade already lands in this PR's own review queue, bound to their own existing record, and that approving it returns their original account intact. The feature you built *is* the migration path, and it's a better one than hand-written SQL. Details inline on `detectLegacyFederatedUsers`.

Everything else below is non-blocking.

tylerpotts:
A few observations that aren't tied to specific lines.

**Scope:** `frontend/src/pages/Login.tsx` has some unrelated churn beyond the revert. The three-step RFC 6749 §4.1 flow comment was rewritten into a shorter paraphrase and an em-dash became a hyphen. Neither relates to identity review, and it adds review surface to a security-sensitive file. Worth reverting to keep the diff honest.

**Docs:** two gaps worth closing in this PR or a quick follow-up.

1. `AGENTS.md`, under "Backend layers": the rule that every externally-authenticated login funnels through `auth.findOrCreateFederatedUser` and that users are matched only by `(issuer, subject)`, never by username or email, is exactly the kind of invariant that gets re-broken by the next auth provider someone adds. It should be written down next to the layer description, along with the note that clients should check `auth.FederatedIdentityReviewErrorCode` before falling through to a generic 401.
2. `docs/docs/ui.md` has a "Groups (Admin)" section but nothing for the new page. Worth a short "Identity Reviews (Admin)" section covering what lands in the queue, what each column means, that Approve permanently binds that issuer/subject to that account, and that Reject currently has no undo in the UI.

**Testing:** the backend tests are genuinely good. The gap is that nothing covers the sequence "review filed, then the same external identity logs in again with different claims." That's the case behind the blocker on `recordFederatedIdentityReview`, and a test asserting a pending review's target can't move would lock the fix in.

Nice work on this overall. The security model is right, and once the approval step is bound to what the admin actually reviewed, the whole thing holds together.

internal/auth/federated_identity.go
MUFFANUJ: This seems less likely to happen, not that it cant happen hm lets address this

MUFFANUJ: Removing the update part is easier/better than putting more effort into making it real-time updates on the admin side.

internal/db/db.go
MUFFANUJ: Since we are dropping this, this concern is gone.

internal/service/admin.go
MUFFANUJ: Let's do this in a follow-up!

internal/auth/federated_identity.go
MUFFANUJ: I will try to keep a note for it if user counts grows.

tylerpotts (commented):
## Review: federated identity review flow

Thanks for this, the shape of the flow (gate the link, record a review, let an admin decide) is right. There are a handful of issues I think need resolving before merge, plus some smaller cleanups. 15 inline comments below; summarising the ones I'd treat as blocking:

**Blocking**

1. `ListFederatedIdentityReviews` returns the *oldest* 100 rows with no status filter and no pagination, and rejected rows are never deleted. Once 100 review rows exist, no new pending review is ever visible to an admin again. Given the docs tell operators to expect a burst of reviews on upgrade, a 150-user org silently locks out 50 people.
2. Deleting a user is a soft delete, so the `OnDelete:CASCADE` on `federated_identities` never fires and the orphaned identity row permanently blocks that subject from ever logging in or being re-provisioned.
3. The `403 identity_review_pending` responses never reach the user: the frontend axios interceptor only handles 401, so cookie-authenticated clients get broken panels with no explanation. The messaging added to `Login.tsx` is only reachable on the `?error=` redirect path.
4. Rejection is irreversible. There is no un-reject/delete/re-open in the API, service, or UI, and the unique index on `(issuer, subject)` prevents a new review from being recorded. One misclick on a screen designed for bulk click-through is a permanent lockout.
5. `authErrorMessage()` echoes an arbitrary `?error=` query param into the login page banner. Not XSS, but it is a first-party-looking content-injection primitive on the credential-entry page that did not exist before this change.

**Worth fixing**

6, 7, 8 (unconditional per-request `db.Save`, TOCTOU on first login, un-actionable reviews for deleted users) and 9, 10 on the frontend. The rest are cleanups.

Happy to talk through any of these if you disagree with the reading.

internal/service/admin.go
```diff
@@ -302,6 +303,108 @@ func (s *AdminService) ListAuditLogs(userIDFilter, actionFilter string) ([]model
...
+	var reviews []models.FederatedIdentityReview
+	if err := s.db.Preload("User").Order("created_at ASC").Limit(100).Find(&reviews).Error; err != nil {
```
tylerpotts: **Blocking: pending reviews become permanently invisible after 100 rows.**

This is `Order("created_at ASC").Limit(100)` with no status filter and no pagination, and `RejectFederatedIdentityReview` only flips `status` rather than deleting the row. So the oldest-100 window fills with terminal (approved/rejected) rows and never drains.

Once 100 review rows exist, a newly created pending review is never returned here, so it never appears in `/admin/identity-reviews` nor in the dashboard "N identity reviews pending" count. `docs/docs/server-setup.md` tells operators to "expect a temporary burst of identity reviews as legacy users sign in", so this is on the upgrade path, not an edge case: an org with 150 pre-existing OIDC users has 50 of them locked out with no admin-visible recourse.

Compare `ListAuditLogs` on line 290, which correctly uses `Order("timestamp DESC").Limit(100)`.

Suggested fix: order `DESC`, filter to `status = 'pending'` for the actionable list (or at minimum paginate), and give terminal rows a way out.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,316 @@
...
+		return nil, errors.New("federated token has no subject")
+	}
+
+	var identity models.FederatedIdentity
+	err := db.Preload("User").
+		Where("issuer = ? AND subject = ?", claims.Issuer, claims.Subject).
+		First(&identity).Error
+	if err == nil {
+		user := identity.User
+		if user.ID != identity.UserID {
```
tylerpotts: **Blocking: deleting a user permanently and silently locks out their federated identity.**

`AdminService.DeleteUser` does `s.db.Delete(&user)`, which is a GORM soft delete, so the DB-level `OnDelete:CASCADE` on `FederatedIdentity` never fires and the identity row survives.

On the next login for that `(issuer, subject)`:
- lines 56-58 find the identity successfully
- `Preload("User")` excludes the soft-deleted row, so `identity.User` is the zero value
- `user.ID` is `uuid.Nil != identity.UserID`, so this guard returns "federated identity is not linked to an active user"

Callers map that to a generic 401. And because the identity *lookup* succeeded, the create-and-review path is never reached, so no new account is ever provisioned. The user is permanently locked out with an opaque error, and there is no admin UI to clear the orphaned identity row.

Options: hard-delete or explicitly unlink federated identities in `DeleteUser`, or treat "identity found but user gone" as a re-provision case rather than a hard failure.

internal/api/handlers/auth.go
```diff
@@ -110,9 +116,30 @@ func SessionCheck(basicAuth *auth.BasicAuthenticator, proxyAdminGroups string) g
...
+			}
 			c.JSON(http.StatusUnauthorized, gin.H{"error": "no proxy session"})
 			return
 		}
 		c.JSON(http.StatusOK, resp)
 	}
 }
+
+func writeFederatedIdentityReviewJSON(c *gin.Context, err error) bool {
+	code, ok := auth.FederatedIdentityReviewErrorCode(err)
```
tylerpotts, line 131: **Blocking: this 403 never surfaces to the user.**

`BasicAuthenticator.Middleware()` (internal/auth/basic.go:203) and `SessionCheck` both return `403 {"error":"identity_review_pending"}`, but `frontend/src/api/client.ts` only branches on `error.response?.status === 401`. A 403 falls through to `Promise.reject`: no token clear, no redirect to `/login`, and nothing reads the `error` code. Every TanStack query errors and the user sees empty or erroring panels with zero explanation.

That means the user-facing messaging added in `Login.tsx` only fires on the `?error=` full-page redirect path (SessionRedirect / OIDCCallback) and is unreachable for cookie-authenticated clients, which is the gateway/proxy deployment shape.

Either return 401 here so the existing interceptor handles it, or teach the interceptor to recognise 403 + a review error code and redirect to `/login?error=...`.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+		if user != nil {
+			return user, "username", nil
+		}
+	}
+
+	if claims.Email != "" {
+		user, err := userByField(db, "email", claims.Email)
+		if err != nil {
+			return nil, "", err
```
tylerpotts: **Blocking: rejection is permanent and irreversible.**

There is no un-reject, delete, or re-open path anywhere: no DELETE/PATCH endpoint, nothing in the service layer, and `FederatedIdentityReviews.tsx` renders `<p>Rejected</p>` where the buttons would be. The unique index on `(issuer, subject)` also prevents a fresh review from being recorded, and `TestOIDCFindOrCreateUser_RejectedReviewBlocksWithoutCurrentCollision` explicitly asserts this check still fires even after the original collision is gone.

`docs/docs/ui.md` concedes "there is currently no undo action in the UI". Combined with the fact that the upgrade path forces a review for *every* legacy federated user, this makes a bulk click-through screen where one misclick permanently bricks a colleague's SSO account, recoverable only by hand-editing the database.

At minimum: an admin action to delete or reopen a review.

frontend/src/pages/Login.tsx
```diff
@@ -13,10 +13,29 @@ type LoginProps = {
...
+  if (error === 'identity_review_pending') {
+    return 'Your identity link request is pending admin approval. Try again after an admin approves it.';
+  }
+  if (error === 'identity_review_rejected') {
+    return 'Your identity link request was rejected by an admin. Contact an administrator if you believe this is a mistake.';
+  }
+  if (!error || error === 'oauth_failed') {
+    return 'Authentication failed';
+  }
+  return error;
```
tylerpotts: **Blocking: this echoes attacker-controlled text into the login page.**

`authErrorMessage()` falls through to `return error;` for anything it does not recognise, and its input is `searchParams.get('error')` (line 84). So `https://nebi.example.com/login?error=Your%20account%20was%20locked.%20Call%20IT%20at%201-800-555-0199` renders that text verbatim inside the styled destructive banner, directly above the real credential form.

React escapes HTML so this is not XSS, but it is a first-party-looking content-injection / phishing primitive on the credential-entry page, and it did not exist before this diff: the previous code hardcoded "Authentication failed".

Fix: return a constant for unrecognised codes.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,316 @@
...
+		if user.ID != identity.UserID {
+			return nil, errors.New("federated identity is not linked to an active user")
+		}
+		if user.AvatarURL != claims.AvatarURL {
+			if err := db.Model(&user).Update("avatar_url", claims.AvatarURL).Error; err != nil {
+				return nil, fmt.Errorf("failed to update federated user profile: %w", err)
+			}
+			user.AvatarURL = claims.AvatarURL
+		}
+		if err := updateFederatedIdentityProfile(db, &identity, claims); err != nil {
```
tylerpotts: **Turns a read-only auth check into a per-request write.**

`updateFederatedIdentityProfile` unconditionally calls `db.Save(identity)` on every successful federated resolution, and `BasicAuthenticator.Middleware()` calls this path on *every* cookie-authenticated request. `Save` writes all columns plus `updated_at` even when nothing changed.

Worse, `identity` was loaded with `Preload("User")` on line 57, so GORM's default association auto-save also emits an `INSERT ... ON CONFLICT DO NOTHING` for the users row on each `Save`.

At 20 req/s that is roughly 2M writes/day of pure churn, which matters against SQLite (single writer). The code this replaced only wrote when the avatar actually changed (`if user.AvatarURL != claims.Picture`).

Fix: compare the profile fields and skip the `Save` when unchanged, and use `Select(...)` to avoid the association write.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,316 @@
...
+func findOrCreateFederatedUser(db *gorm.DB, claims federatedUserClaims) (*models.User, error) {
+	claims = normalizeFederatedClaims(claims)
+	if claims.Issuer == "" {
+		return nil, errors.New("federated token has no issuer")
+	}
+	if claims.Subject == "" {
+		return nil, errors.New("federated token has no subject")
+	}
+
+	var identity models.FederatedIdentity
```
tylerpotts: **TOCTOU: this lookup is outside the transaction that creates the user and identity.**

Two concurrent first logins for the same brand-new identity (two tabs, or a client retry) both reach here, both get `gorm.ErrRecordNotFound`, and both enter the `db.Transaction` on line 78. The second insert violates `idx_federated_identity_issuer_subject` and returns "failed to create federated identity: UNIQUE constraint failed".

That error is not a review sentinel, so `OIDCCallback` redirects to `/login?error=oauth_failed` and the user sees "Authentication failed" on what should have been a successful first login.

Fix: move the lookup inside the transaction, or use `clause.OnConflict` and re-read on conflict.

internal/service/admin.go
```diff
@@ -302,6 +303,108 @@ func (s *AdminService) ListAuditLogs(userIDFilter, actionFilter string) ([]model
...
+	var identity models.FederatedIdentity
+	if err := s.db.Transaction(func(tx *gorm.DB) error {
+		var review models.FederatedIdentityReview
+		if err := tx.Preload("User").First(&review, "id = ?", reviewID).Error; err != nil {
+			if errors.Is(err, gorm.ErrRecordNotFound) {
+				return ErrNotFound
+			}
+			return err
+		}
+		if review.User.ID != review.UserID {
```
tylerpotts, line 384: **A review whose user is later deleted becomes un-actionable and reports a misleading 404.**

Flow: legacy user logs in post-upgrade, pending review created bound to their ID. Admin deletes the user before reviewing (soft delete). Admin then clicks Approve: line 322 loads the review fine, but `Preload("User")` yields a zero User, so this check returns `ErrNotFound` and `handleServiceError` renders 404 "Not found" for a review that is demonstrably right there in the table.

`recordFederatedIdentityReview` (federated_identity.go:186-193) never rebinds an existing pending review either, so the stale row is immortal, and it consumes one of the 100 slots from the pagination issue in `ListFederatedIdentityReviews`.

Fix: return a `ValidationError`/`ConflictError` naming the deleted user so the message is actionable, and add a way to discard the review.

frontend/src/pages/admin/AdminDashboard.tsx
```diff
@@ -108,18 +125,29 @@ export const AdminDashboard = () => {
...
   ).length;
+  const pendingIdentityReviews = displayedIdentityReviews.filter(
+    isPendingFederatedIdentityReview,
+  ).length;
 
   const remoteQueries = [
     remoteWorkspacesQuery,
     remoteJobsQuery,
     remoteStatsQuery,
+    remoteReviewsQuery,
```
tylerpotts: **This makes the whole dashboard report "remote unreachable" against any server that predates this PR.**

`useRemoteFederatedIdentityReviews` hits `GET /remote/admin/federated-identity-reviews`. Against an older server, `RemoteHandler.ListAdminFederatedIdentityReviews` gets a 404 from cliclient and converts it to `http.StatusBadGateway`, so `query.isError` is true. Line 152's `remoteQueries.some((query) => query.isError)` then renders `RemoteUnreachableBanner` across the entire dashboard even though workspaces, jobs, and stats all loaded fine.

There is no capability or version check on this endpoint. Either exclude the reviews query from the unreachable computation, or have the proxy translate a 404 into "feature not supported" rather than 502.

frontend/src/pages/admin/FederatedIdentityReviews.tsx
```diff
@@ -0,0 +1,312 @@
...
+  );
+};
+
+export const FederatedIdentityReviews = () => {
+  const { data: reviews, isLoading: reviewsLoading } =
+    useFederatedIdentityReviews();
+  const approveReview = useApproveFederatedIdentityReview();
+  const rejectReview = useRejectFederatedIdentityReview();
+
+  const isLocalMode = useModeStore((s) => s.isLocalMode());
```
tylerpotts: **This should use `useRemoteView()` and render `RemoteUnreachableBanner`.**

The repo-root `AGENTS.md` (Frontend section) says new or updated pages should gate remote data and the unreachable banner with `useRemoteView()` from `src/hooks/useRemote.ts`, and that the remaining hand-rolled pages are a backlog tracked in https://github.com/nebari-dev/nebi/issues/504. Lines 52-56 re-derive `isLocalMode` / `viewMode` / `useRemoteServer()` / `isRemoteConnected` / `shouldShowRemote` by hand, adding a new file to that backlog.

The concrete consequence: with `viewMode === 'remote'` and the remote server down, `remoteReviews` is undefined, `displayedReviews` falls back to `[]` on line 88, and the page renders "No identity reviews", telling the admin there is nothing to approve when the fetch actually failed. Every other remote-aware page shows the banner instead.

internal/api/router.go
```diff
@@ -344,6 +349,9 @@ func NewRouter(cfg *config.Config, db *gorm.DB, q queue.Queue, exec executor.Exe
 				remote.GET("/admin/registries", remoteHandler.ListAdminRegistries)
 				remote.GET("/admin/audit-logs", remoteHandler.ListAdminAuditLogs)
 				remote.GET("/admin/dashboard/stats", remoteHandler.GetAdminDashboardStats)
+				remote.GET("/admin/federated-identity-reviews", remoteHandler.ListAdminFederatedIdentityReviews)
+				remote.POST("/admin/federated-identity-reviews/:id/approve", remoteHandler.ApproveAdminFederatedIdentityReview)
```
tylerpotts: **These mutating admin proxies are outside `RequireAdmin`.**

Every server-side admin route sits in `admin := protected.Group("/admin"); admin.Use(middleware.RequireAdmin(localMode, rbacProvider))` (lines 275-276), but these two sit in `protected.Group("/remote")`, which has no such middleware. They are also the first *mutating* remote/admin proxies, since all the prior ones are GETs.

In practice local mode is single-user desktop and the remote server re-checks admin on the token, so the exposure is bounded today. But it is a defence-in-depth gap that widens the moment the remote group is reachable in a multi-user context, and it is inconsistent with the gating applied to the identical operations 60 lines above.

frontend/src/pages/admin/FederatedIdentityReviews.tsx
```diff
@@ -0,0 +1,312 @@
...
+                      {formatDate(review.created_at)}
+                    </td>
+                    <td className="p-4">
+                      {isPendingFederatedIdentityReview(review) ? (
+                        <div className="flex justify-end gap-2">
+                          <Button
+                            size="sm"
+                            variant="outline"
+                            onClick={() => setReviewToReject(review)}
+                            disabled={isMutating}
```
tylerpotts: **A single global `isMutating` spins every row's buttons.**

`isMutating` (lines 78-82) ORs all four mutation `isPending` flags, so approving one review disables and shows a spinner on all Approve *and* Reject buttons in the table. With 30 pending reviews that is 60 buttons spinning, and the admin cannot tell which review is actually in flight or queue a second action.

TanStack's `mutation.variables` already carries the in-flight `reviewId`. Compare it to `review.id` so only the affected row spins, and gate `disabled` per row.

internal/auth/local.go
```diff
@@ -25,9 +25,11 @@ func NewLocalAuthenticator(db *gorm.DB) (*LocalAuthenticator, error) {
 	err := db.Where("username = ?", localUsername).First(&user).Error
 	if errors.Is(err, gorm.ErrRecordNotFound) {
 		user = models.User{
-			Username:     localUsername,
-			Email:        localUsername + "@nebi.local",
-			PasswordHash: "-", // no password; local mode never checks credentials
+			Username: localUsername,
+			Email:    localUsername + "@nebi.local",
+			// Keep non-empty: migrations use empty password hashes to find
```
tylerpotts: **This comment points at a migration that does not exist on this branch.**

The comment says migrations use empty password hashes to find legacy federated users needing issuer/subject backfill, but `detectLegacyFederatedUsers` was added mid-PR and removed again before the final commit, and nothing in the tree queries `password_hash = ''`.

So the comment asserts a load-bearing invariant with no enforcer. A future contributor either preserves a constraint for no reason or goes looking for the migration and cannot find it. Either drop it or restore the original "no password; local mode never checks credentials".

internal/api/handlers/auth.go
```diff
@@ -30,6 +30,9 @@ func Login(authenticator auth.Authenticator) gin.HandlerFunc {
 
 		resp, err := authenticator.Login(req.Username, req.Password)
 		if err != nil {
+			if writeFederatedIdentityReviewJSON(c, err) {
```
tylerpotts: **This branch is unreachable.**

`Login(authenticator auth.Authenticator)` is wired with either `LocalAuthenticator` (returns the fixed local user, no error) or `BasicAuthenticator.Login` (bcrypt check against `users`, returns `ErrInvalidCredentials`). Neither ever calls `findOrCreateFederatedUser`, so `auth.FederatedIdentityReviewErrorCode(err)` can never return `ok == true` here.

The cost is a reader inferring that username/password login participates in the review flow, plus a branch no test can exercise. The genuinely reachable call sites are `SessionCheck`, `DeviceToken`, and `Middleware`.

frontend/src/pages/Login.tsx
```diff
@@ -13,10 +13,29 @@ type LoginProps = {
...
+};
+
+const authErrorTone = (error: string) =>
+  error === 'identity_review_pending' ? 'warning' : 'destructive';
+
 export const Login = ({ isDarkMode }: LoginProps) => {
   const [username, setUsername] = useState('');
   const [password, setPassword] = useState('');
   const [error, setError] = useState('');
+  const [errorTone, setErrorTone] = useState<'destructive' | 'warning'>(
```
tylerpotts: **`errorTone` is derivable state.**

`setError` and `setErrorTone` are paired at lines 84-85, 101-102, 114-115, and 126-127, and line 115 exists purely to undo a stale warning tone from a previous render. Any future `setError` that forgets its `setErrorTone` partner leaves an amber "pending" banner wrapped around an unrelated failure.

Storing the error *code* (or null) in one state and computing `authErrorMessage(code)` / `authErrorTone(code)` at render time removes the second `useState` and all four pairings, and makes the mismatched state unrepresentable.

internal/service/admin.go
MUFFANUJ: I was on track to introduce pagination in a follow-up. For now, with this being a blocker, maybe I will address it here.

docs/docs/ui.md
```diff
@@ -35,3 +35,9 @@ Admins can manage **groups** to grant workspace, registry, and admin access to m
 - **Grant access**: share a workspace with a group via the workspace's Share dialog (User/Group toggle). Registry and admin grants for groups are admin-only operations.
 
 OIDC groups (where the `groups` claim in the user's ID token creates them automatically) display with a blue "oidc" badge and are read-only in the UI.
+
+## Identity Reviews (Admin)
+
+Admins can review blocked external identity links from **Admin -> Identity Reviews**. A review appears when an OIDC, proxy-auth, or device-flow login presents an issuer/subject pair that is not yet bound to a Nebi user but its username or email claim collides with an existing account.
+
+Each row shows the target user, issuer, subject, collision field, profile claims, status, and review time. **Approve** permanently binds that issuer/subject to the shown account and future logins use that binding even if username or email claims later change. **Reject** blocks that issuer/subject from linking to the account; there is currently no undo action in the UI.
```
tylerpotts: **Blocker:** this sentence is now false: "Reject blocks that issuer/subject from linking to the account; there is currently no undo action in the UI." This commit adds the undo - the Rejected tab's Discard action, whose confirm dialog even documents the behavior ("The next OAuth login for this identity can create a fresh pending review"). Please rewrite to describe the Pending/Rejected tabs and Discard. This is the only thing between this PR and approval.

internal/service/admin.go
```diff
@@ -175,13 +176,24 @@ func (s *AdminService) DeleteUser(userID uuid.UUID, adminUserID uuid.UUID) error
...
 	}
 
-	if err := s.db.Delete(&user).Error; err != nil {
-		return fmt.Errorf("delete user: %w", err)
-	}
+	if err := s.db.Transaction(func(tx *gorm.DB) error {
+		if err := tx.Unscoped().Where("user_id = ?", user.ID).Delete(&models.FederatedIdentity{}).Error; err != nil {
+			return fmt.Errorf("delete federated identities: %w", err)
+		}
+		if err := tx.Unscoped().Where("user_id = ?", user.ID).Delete(&models.FederatedIdentityReview{}).Error; err != nil {
```
tylerpotts, line 187: **Question:** this deletes reviews with no status filter, rejected included. Consequence: delete user X, and an identity whose link to X an admin explicitly *rejected* is no longer blocked - `enforceRejectedFederatedIdentityReview` finds nothing and the next login auto-provisions a fresh account. The reject audit row survives, and "the rejection was about that collision" is a defensible reading, but it quietly reverses the earlier change that made rejection durable beyond the collision. If un-banning is intended, a comment here saying so would settle it; if not, filter this delete to pending.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,415 @@
...
+		AvatarURL:      claims.AvatarURL,
+		Status:         models.FederatedIdentityReviewStatusPending,
+	}
+	if err := db.Create(&review).Error; err != nil {
+		return fmt.Errorf("failed to record federated identity review: %w", err)
+	}
+	return nil
+}
+
+func deleteStalePendingFederatedIdentityReview(db *gorm.DB, claims federatedUserClaims) error {
```
tylerpotts, line 294: **Worth fixing:** this hard-delete runs on behalf of an unauthenticated login and leaves no trace: no audit row, no log line (the file's only `slog` is user creation). Scenario: an attacker's collision files a review the admin can see; the attacker changes `preferred_username` at the IdP and logs in again; the evidence of the attempted account-claim vanishes mid-investigation and a fresh account is provisioned. No privilege is gained, which is why this is not a blocker, but at minimum `slog.Info` the deletion with issuer/subject and the old collision target; better, an audit row or a `superseded` status instead of deletion.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,415 @@
...
+
+func federatedEmailWithSuffix(email, suffix string, n int) string {
+	local, domain, ok := strings.Cut(email, "@")
+	if !ok || local == "" || domain == "" {
+		return fmt.Sprintf("%s-%s-%d@nebi.local", email, suffix, n)
+	}
+	return fmt.Sprintf("%s+%s-%d@%s", local, suffix, n, domain)
+}
+
+func isUniqueConstraintError(err error) bool {
```
tylerpotts, line 442: **Worth fixing:** the `"constraint failed"` clause over-matches - on this repo's SQLite driver, NOT NULL and CHECK violations also contain it, while `"unique constraint"` alone already matches both SQLite and Postgres unique messages. Unreachable today (the User model can't produce non-unique constraint errors on this path), but latent. The sharper half: when a retry triggers, the underlying error is discarded, so retry exhaustion logs only "retry federated identity resolution" with the actual conflict unrecoverable. Suggest dropping the extra clauses and returning `fmt.Errorf("%w: %v", errRetryFederatedIdentityResolution, err)` so `errors.Is` still matches but the cause survives into logs.

internal/models/federated_identity.go
```diff
@@ -0,0 +1,75 @@
...
+	Username      string         `json:"username"`
+	Email         string         `json:"email"`
+	EmailVerified bool           `json:"email_verified"`
+	Name          string         `json:"name"`
+	AvatarURL     string         `json:"avatar_url"`
+	CreatedAt     time.Time      `json:"created_at"`
+	UpdatedAt     time.Time      `json:"updated_at"`
+	DeletedAt     gorm.DeletedAt `gorm:"index" json:"-"`
```
tylerpotts: **Worth fixing (latent):** both new models carry `gorm.DeletedAt` alongside hard unique indexes on `(issuer, subject)`. All six current delete sites correctly use `Unscoped()`, so this can't fire today - but I verified empirically that a soft-deleted identity row is invisible to the scoped lookup while still winning the `OnConflict`, which means the first future scoped `Delete` permanently bricks that identity's login. The models are new in this PR, so there's no migration cost: drop the `DeletedAt` fields, or comment that scoped deletes are forbidden on these tables.

frontend/src/pages/admin/AuditLogs.tsx
```diff
@@ -16,6 +16,9 @@ const ACTION_COLORS: Record<string, string> = {
   revoke_permission: 'bg-orange-100 text-orange-800 border-orange-300',
   make_admin: 'bg-purple-100 text-purple-800 border-purple-300',
   revoke_admin: 'bg-zinc-100 text-zinc-800 border-zinc-300',
+  approve_federated_identity:
+    'bg-emerald-100 text-emerald-800 border-emerald-300',
+  reject_federated_identity: 'bg-red-100 text-red-800 border-red-300',
```
tylerpotts, line 21: **Worth fixing:** `discard_federated_identity_review` is missing from `ACTION_COLORS` and the action filter dropdown, while `approve_federated_identity` / `reject_federated_identity` - added in this same PR - appear in both. Discard events render with the generic fallback badge and can't be filtered, which undercuts the audit integration this PR adds. Two-line fix. Related nit: the constant breaks the sibling naming scheme (`..._identity` vs `..._identity_review`), which makes log queries irregular; all three target the same resource.

internal/api/router.go
```diff
@@ -345,10 +351,18 @@ func NewRouter(cfg *config.Config, db *gorm.DB, q queue.Queue, exec executor.Exe
 				remote.GET("/jobs", remoteHandler.ListJobs)
 
 				// Admin proxies (for view mode toggle in admin pages)
-				remote.GET("/admin/users", remoteHandler.ListAdminUsers)
-				remote.GET("/admin/registries", remoteHandler.ListAdminRegistries)
-				remote.GET("/admin/audit-logs", remoteHandler.ListAdminAuditLogs)
-				remote.GET("/admin/dashboard/stats", remoteHandler.GetAdminDashboardStats)
+				remoteAdmin := remote.Group("/admin")
+				remoteAdmin.Use(middleware.RequireAdmin(localMode, rbacProvider))
```
tylerpotts, line 358: Nit: this middleware can never enforce anything - the `/remote` group only mounts when `localMode` is true, and `RequireAdmin` unconditionally skips in local mode. Real enforcement is the remote server rejecting the proxied token. It's the defense-in-depth consistency I asked for, so keep it, but add a comment saying it's intentionally symbolic so nobody reads protection into it.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,415 @@
...
+	if errors.Is(err, gorm.ErrRecordNotFound) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("failed to check federated identity review: %w", err)
+	}
+	return fmt.Errorf("%w: issuer %s subject %s", errFederatedIdentityRejected, claims.Issuer, claims.Subject)
+}
+
+func recordFederatedIdentityReview(db *gorm.DB, userID uuid.UUID, field string, claims federatedUserClaims) error {
```
tylerpotts: Nit: two concurrent first logins for the same identity can both pass the `First` check here; the loser's `Create` hits the review unique index and surfaces as `oauth_failed` instead of `identity_review_pending`, once, self-healing on retry. Routing that error through `isUniqueConstraintError` and treating it as review-pending would close it.

tylerpotts (changes requested):
## Re-review at 51bd4cd

This round closes it out. All 15 items from my last review are genuinely fixed, and I verified them in a worktree rather than trusting the replies: three of the fixes I checked by mutation (broke the fix, watched the new test go red at the exact assertion, restored it). Specifics worth calling out:

- **List visibility (was blocker 1):** status filter with pending default, `created_at DESC`, validated statuses, plumbed through handler, cliclient, remote proxy, hooks, and the new Pending/Rejected tabs. There is still no pagination past 100 *pending* rows and the dashboard tile caps at 100, but that is a far narrower gap than "pending reviews permanently invisible", and a fine follow-up.
- **Soft-delete orphan lockout (was blocker 2):** fixed twice, belt and suspenders. `DeleteUser` hard-deletes identities and reviews in a transaction, and `loadFederatedIdentityUser` clears orphaned rows and re-provisions. Removing the `Unscoped()` turns `TestAdminDeleteUser` red; breaking the orphan cleanup turns the new re-provision test red.
- **403 surfacing (was blocker 3):** the interceptor now handles the two review codes and redirects with the code intact. I traced the loop risk: nothing calls `/auth/session` through axios, and `Login.tsx` returns on `?error=` before its session check, so no redirect loop.
- **Un-reject (was blocker 4):** the Discard endpoint is the right shape: hard delete (I confirmed empirically that a soft delete would still occupy the unique index and re-brick the identity, so the `Unscoped()` is load-bearing), audit-logged in the transaction, admin-gated, mirrored in cliclient and the remote proxy, and the next login files a fresh review.
- **Error echo (was blocker 5):** closed switch, constant default, and the test asserts the injected string is not rendered. `searchParams` appears nowhere else in the frontend.
- The TOCTOU fix is correct on both drivers: I verified on SQLite that `OnConflict DoNothing` yields `RowsAffected == 0` and the sentinel rolls back the loser's provisional user row; Postgres by analysis. The per-request write churn fix is pinned by the new `updated_at` test, which does catch the regression (GORM's `Select().Updates` bumps `updated_at` even for identical values).
- The per-row `mutation.variables` spinner pattern is the first of its kind in this codebase and a strict improvement over the whole-table disables elsewhere.

I re-ran everything at this commit: `go build`/`go vet`/`go test ./internal/...` green, `npx vitest run` 32 files / 219 tests green, and `swag init` produces no diff.

**One blocker, and it is a one-line doc fix:** `docs/docs/ui.md` line 43 still says "there is currently no undo action in the UI", which this commit makes false - the Rejected tab's Discard button is exactly that undo, and its own confirm dialog describes the semantics. An admin who trusts the doc will resort to DB surgery for something the UI now does. Please update that sentence (and ideally mention the Pending/Rejected tabs) and this is an approve.

**One question I'd like answered before merge** (inline on `DeleteUser`): deleting a user now hard-deletes their *rejected* reviews too, which silently un-bans an identity an admin explicitly rejected. Maybe intended, since the rejection was collision-specific, but it cuts against the earlier change that made rejection durable beyond the collision. Your call, but say which semantics you mean.

Non-blocking but worth folding in while you are here, details inline: the new `discard_federated_identity_review` audit action is invisible in the AuditLogs UI (missing from the filter dropdown and color map, unlike its approve/reject siblings from this same PR); `deleteStalePendingFederatedIdentityReview` destroys the only trace of an attempted collision with no log or audit row; `isUniqueConstraintError` over-matches on SQLite and discards the underlying error on retry exhaustion; and both new models carry a `DeletedAt` that must never be used given the hard unique indexes - every current delete site correctly uses `Unscoped()`, but the first future scoped delete quietly re-creates the bricked-identity bug.

Checklist note: "non-breaking" is now accurate with the migration guard gone.

MUFFANUJ:
Thanks @tylerpotts, I will push a commit to address these. I looked over it again and want to address some more new stuff.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,453 @@
...
+		Username:                claims.PreferredUsername,
+		Email:                   claims.Email,
+		EmailVerified:           claims.EmailVerified,
+		Name:                    claims.Name,
+		AvatarURL:               claims.AvatarURL,
+		Status:                  models.FederatedIdentityReviewStatusPending,
+	}
+	if err := db.Create(&review).Error; err != nil {
+		if isUniqueConstraintError(err) {
+			// A concurrent first login may have created this review after the
```
tylerpotts, line 285: **Suggestion:** this swallow works on SQLite but not on Postgres. The Create runs inside `db.Transaction`; on Postgres the failed INSERT aborts the transaction, so when this returns nil and the callback commits, pgx returns `ErrTxCommitRollback` (pgx v5.9.2 tx.go:90, propagated by gorm finisher_api.go's `tx.Commit()`), and the losing login still surfaces as a generic failure instead of `identity_review_pending`. Not a regression (the pre-fix code failed the same way), but the comment's promise only holds on SQLite. Concrete fix that works on both drivers: `tx.Clauses(clause.OnConflict{DoNothing: true}).Create(&review)` like the identity insert above already does, then treat `RowsAffected == 0` as the existing-review case.

internal/service/admin.go
```diff
@@ -175,13 +176,28 @@ func (s *AdminService) DeleteUser(userID uuid.UUID, adminUserID uuid.UUID) error
...
-	}
+	if err := s.db.Transaction(func(tx *gorm.DB) error {
+		if err := tx.Unscoped().Where("user_id = ?", user.ID).Delete(&models.FederatedIdentity{}).Error; err != nil {
+			return fmt.Errorf("delete federated identities: %w", err)
+		}
+		// Reviews are collision-specific to this user. Once the target user is
+		// deleted, remove both pending and rejected reviews so future sign-ins
+		// for the same issuer/subject are evaluated against the remaining
+		// active accounts.
+		if err := tx.Unscoped().Where("user_id = ?", user.ID).Delete(&models.FederatedIdentityReview{}).Error; err != nil {
```
tylerpotts, line 187: **Suggestion:** this only deletes reviews where `user_id` matches, so deleting the *email-matched* user of an ambiguous review leaves `collision_email_user_id` dangling. `HasAmbiguousCollision` compares the stored IDs with no existence check, nothing ever refreshes the collision columns (subsequent logins keep an existing pending review as-is, by design), so the review stays permanently unapprovable even though the admin resolved the conflict the natural way. The only way out is reject then discard then re-login. Suggest also cleaning up (or nulling the collision column of) reviews referencing the deleted user via `collision_username_user_id` / `collision_email_user_id`, which would also make the audit trail account for those vanished reviews.

internal/service/admin.go
```diff
@@ -324,6 +340,176 @@ func (s *AdminService) ListAuditLogs(userIDFilter, actionFilter string) ([]model
...
+		if review.User.ID != review.UserID {
+			return &ConflictError{
+				Message: fmt.Sprintf("federated identity review is linked to deleted user %s; discard it and ask the user to sign in again", review.UserID),
+			}
+		}
+		if !review.IsPending() {
+			return &ConflictError{Message: "federated identity review was rejected"}
+		}
+		if review.HasAmbiguousCollision() {
+			return &ConflictError{Message: "federated identity review has username and email collisions with different users; discard it after resolving the conflicting accounts"}
```
tylerpotts, line 393: **Suggestion:** this advice isn't executable from the UI. The API does allow discarding a pending review, but the frontend renders Discard only on non-pending rows and ui.md documents Discard only on the Rejected tab, so a UI admin facing an ambiguous pending review sees a disabled Approve and no Discard. Pick one workflow: either reword this to "reject it, then discard it from the Rejected tab", or render Discard for ambiguous pending rows (and document the ambiguous state in ui.md either way; "multiple users" and the disabled Approve currently appear in no doc).

docs/docs/server-setup.md
```diff
@@ -86,6 +86,16 @@ When OIDC authentication is configured, nebi requests the `groups` scope alongsi
...
 
+### Legacy federated identity migration
+
+Nebi versions that predate issuer/subject identity binding may have users created by OIDC, proxy auth, or device flow with no row in `federated_identities`. Nebi now treats the first post-upgrade login for those users like any other external-identity collision when the incoming username or verified email still matches an existing account: it creates a pending **Identity Review** for an admin to approve.
+
+After upgrading, expect a temporary burst of identity reviews as legacy users sign in. Approving a review binds that external issuer/subject to the existing Nebi account and audit-logs the decision. Rejecting a review blocks that external identity from linking to the account; admins can later discard the rejected review from the **Rejected** tab to allow a fresh review on the next login.
```
tylerpotts, line 93: **Suggestion (docs):** worth an explicit operator warning here that email collision detection requires the IdP to send `email_verified: true`. IdPs that omit the claim (Azure AD is the classic case, and many proxy-auth setups) are now treated as unverified everywhere: users get synthetic `@nebi.local` emails and email-only legacy matches never file a review, silently creating duplicate accounts, which is exactly the migration case this section is about. The gating itself is right; whether a config knob to trust unverified IdP emails is wanted can be a follow-up decision.

internal/auth/federated_identity.go
```diff
@@ -0,0 +1,453 @@
...
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("load stale federated identity review: %w", err)
+	}
+	if err := db.Unscoped().Delete(&review).Error; err != nil {
+		return fmt.Errorf("delete stale federated identity review: %w", err)
+	}
+	slog.Info(
+		"Deleted stale pending federated identity review after claims no longer collide",
```
tylerpotts, line 310: Nit: this logs even when the Delete-by-PK affected 0 rows (row removed by a concurrent login between the `First` and the `Delete`; GORM returns no error for that). Check `RowsAffected` before logging so the log can't claim a deletion that didn't happen.

frontend/src/pages/admin/FederatedIdentityReviews.tsx
```diff
@@ -0,0 +1,480 @@
...
+
+type ReviewStatusTab = Exclude<FederatedIdentityReviewStatusFilter, 'all'>;
+
+const reviewDisplayName = (review: FederatedIdentityReview) =>
+  review.username || review.name || review.email || review.subject;
+
+const formatDate = (value: string) =>
+  value ? new Date(value).toLocaleString() : '—';
+
+const hasAmbiguousCollision = (review: FederatedIdentityReview) =>
```
tylerpotts, line 45: Nit: `isPendingFederatedIdentityReview`, the frontend mirror of the model's other status predicate, lives in `types/models.ts` and is imported here, while its new sibling is page-local. Suggest moving `hasAmbiguousCollision` next to it so future consumers (e.g. the dashboard pending card) don't re-derive it.

tylerpotts (approved):
## Re-review at 25d8a5d

Everything from my last review is addressed, and I verified it at this commit rather than trusting replies: `go build` / `go vet` / `go test ./internal/...` green, `npx vitest run` 32 files / 220 tests green, `swag init` produces no diff, and two of the fixes I checked by mutation (broke the fix, watched the new test go red at the exact assertion, restored it).

**Dispositions from my 08-18 review:**
- **ui.md blocker: fixed.** The false "no undo" sentence is gone and every claim in the rewritten text matches the code, including the verified-email wording.
- **DeleteUser question: answered in code.** The intent comment above the review delete declares the un-ban-on-user-deletion semantics deliberately. Good enough for me.
- **Stale-review deletion trace: fixed** (slog with review id, user, issuer, subject, collision field).
- **isUniqueConstraintError: fixed and mutation-verified.** Re-adding the old `constraint failed` clause turns `TestIsUniqueConstraintErrorOnlyMatchesUniqueConstraints` red at exactly the NOT NULL assertion. Wrapping the cause with `%w: %v` preserves it for logs.
- **DeletedAt: fixed.** Dropped from both models; nothing in the repo depends on soft delete for these tables and the fields never shipped, so no migration cost.
- **AuditLogs UI + naming: fixed.** Discard is in the color map and filter dropdown, and the renamed `discard_federated_identity` constant matches its siblings with zero stale references repo-wide.
- **Symbolic RequireAdmin comment: added.**
- **Concurrent-first-login nit: fixed on SQLite**, but not on Postgres, see the inline comment. Not a regression, so still non-blocking.

**The unprompted new work is good.** The verified-email gating is correct hardening, applied uniformly at the single choke point all three auth entry points funnel through, and no path stores an unverified claim email in `User.Email` (all four user-creation sites checked). The ambiguous-collision model, service gate, and UI treatment correctly refuse a coin-flip approve, and the reject race fix is pinned by a test I mutation-verified: reverting to the old `Save` code makes `TestAdminRejectFederatedIdentityReviewDoesNotFallbackInsert` fail as designed.

Architecture-wise the delta is clean: the nullable collision FKs match the existing pointer-FK convention, the model-predicate + service-ConflictError gate mirrors `IsPending`, and the inline page components match how other admin pages are structured.

**Approving.** Five non-blocking suggestions inline; the Postgres review-create swallow and the dangling collision IDs on user deletion are the two I'd most like to see land before the next release. One nit not worth an inline thread: the `Unscoped()` calls on the two now-hard-delete tables are no-ops and could be dropped so future readers don't infer soft delete (the ones on `models.User` are still meaningful, keep those).

Not verified: a live Postgres reproduction of the transaction-abort finding (no instance available); the claim rests on gorm v1.31.0 / pgx v5.9.2 sources I read in the module cache.

MUFFANUJ:
Let's merge this; the suggestion will be followed up in a new PR

PR #472, opened 2026-07-28
Wire remote registry creation through the local/remote toggle

### LINKED ISSUE #468: [BUG] - Can't add remote OCI registries
### Description

I'm using nebi inside of JHub on Nebari.

If you click on "Remote" and add a registry, it creates the registry as a "local" registry. I expected to create a remote registry if I'm on the remote toggle. 

Is it even possible to create a remote registry?

Is it possible for platform admins to set an OCI registry for all users (even if its local)?

### Steps to reproduce

1. Click the remote toggle
2. Click add registry
3. Add the registry
4. Observe that the registry is only under local

### Expected behavior

I expected to create a remote registry if I was on the remote toggle. 

### Environment

_No response_

### Additional context

_No response_

aktech:
Add Registry always POSTed to the local admin registries endpoint regardless of the Local/Remote toggle, since no proxy route existed for creating a registry on a connected remote server (only GET list proxies did). Fixes #468.

- Add RemoteHandler.CreateAdminRegistry proxy + POST /remote/admin/registries route, mirroring the existing workspace local/remote create pattern
- CreateRegistryDialog now takes an isRemote prop and posts to the remote endpoint when the Remote toggle is active
- Add missing api_token field to cliclient.CreateRegistryRequest so it isn't silently dropped on the proxied path

Demo

---
https://github.com/user-attachments/assets/e4fbd85f-9c1c-4bf3-90dd-b59408267806


aktech:
A better way to do this would be fixing this https://github.com/nebari-dev/nebi/issues/514, but I am skipping that for now to limit scope.

internal/cliclient/types.go
```diff
@@ -100,14 +100,17 @@ type WorkspaceVersion struct {
...
-	Restricted bool   `json:"restricted"`
-	CreatedAt  string `json:"created_at"`
-	UpdatedAt  string `json:"updated_at"`
+	ID            string `json:"id"`
+	Name          string `json:"name"`
+	URL           string `json:"url"`
+	Username      string `json:"username,omitempty"`
+	IsDefault     bool   `json:"is_default"`
+	Namespace     string `json:"namespace,omitempty"`
+	HasAPIToken   bool   `json:"has_api_token"`
```
pmeier, line 109: Should this be 

```suggestion
	HasCredentials   bool   `json:"has_credentials"`
```

given that it could also be a password?

pmeier (approved):

PR #477, opened 2026-07-30
feat: admin-provisioned OCI registries via config.yaml

### LINKED ISSUE #475: [FEAT] -  Add ability for admins to set OCI Registry for all users
### Problem

I'd like for a nebi admin to be able to set OCI registries for all users. The OCI registry that I want to add will require a username/password but everyone on my platform has access to this registry with their own credentials so they are allowed access. 

Related - I'd like for the ability to _remove_ the current default OCI registry. 

Note: I'm interested in the Nebi experience inside of JHub. I don't have the standalone nebi pack installed, just what comes with the data science pack. 

### Proposed solution

It would be best if this was a yaml configuration value that would persist on redeploy.

I'm not sure the best way to handle credentials for this. In my current use case, everyone on the platform will be able to access this OCI registry through their credentials so giving them access through one set of keys is an acceptable short term solution, but you should probably think through how best to manage this going forward. Maybe users have to add their own username/password in order to use a private registry?

### Alternatives considered

_No response_

### Additional context

_No response_

tylerpotts:
Closes https://github.com/nebari-dev/nebi/issues/475

## Summary

Admins can now provision OCI registries for all users declaratively, and remove the built-in default registry.

> **Note (2026-08-07):** requirements changed to support **public OCI registries only**. Credential fields (`username`/`password`/`api_token`), `${ENV_VAR}` expansion, and credential encryption were removed from this PR; authenticated registries can still be added per-instance through the admin UI/API/CLI as before.

- New optional `registries:` section in `config.yaml`: a `seed_default` flag plus a list of entries (name, url, namespace, default). Entries are public registries and carry no credentials.
- At boot (both team and local mode), entries are reconciled into the `oci_registries` table in one transaction, keyed by the unique name: created/updated, marked `config_managed`, and removed when dropped from config. Name collisions with user-created registries are taken over by config with a logged warning (any credentials the user-created row had are cleared). The upsert is concurrent-boot safe (savepoint + duplicate-key fallback) for shared-Postgres deployments.
- Config-managed registries are locked: update/delete via API/UI/CLI returns 409, and the default flag cannot be stolen from a config-managed default by creating/updating another registry. The admin UI shows a "Managed" badge and disables edit/delete.
- `registries.seed_default: false` disables seeding of the built-in `quay.io/nebari_environments` registry. Independently, the seed is now one-time (tracked by a marker in a new `system_settings` table), so deleting the default registry via the admin UI finally sticks across restarts. Existing databases get the marker backfilled without being touched.

Companion chart PR for the data-science-pack (mounts this config into JupyterHub user pods): https://github.com/nebari-dev/data-science-pack/pull/202

## Test plan

- [x] Config parsing/validation unit tests
- [x] Reconciler behavior tests: create, update, takeover (including credential clearing), stale removal, remove-then-re-add, default switching (9 cases)
- [x] Seed flag + one-time marker tests, including backfill for pre-existing databases (4 cases)
- [x] Service-level and HTTP-level 409 locking tests, default-steal protection tests
- [x] Frontend component tests for the Managed badge and disabled actions
- [x] End-to-end smoke test: `nebi serve` in local mode with a config-defined registry serves it via `/api/v1/registries` with `config_managed: true`, default seed suppressed; default seed still works with no config
- [x] Full suites green: `go test ./internal/...`, frontend vitest (185 tests)


tylerpotts:
Added a cherry-pick of https://github.com/nebari-dev/nebi/pull/490 (netguard allowed-origins fix, issue https://github.com/nebari-dev/nebi/issues/489) so the pr-477 image is testable behind jupyter-server-proxy. Will drop the commit in a rebase if #490 merges first.

.gitignore
```diff
@@ -83,3 +83,6 @@ docs/.docusaurus/
 # Pixi
 pixi.lock
 .pixi/
+
+# Superpowers working docs (specs/plans)
+docs/superpowers/
```
pmeier: We are not using this. Please remove

config.yaml.example
```diff
@@ -4,6 +4,12 @@
 server:
   port: 8460
   mode: development  # or "production"
+  # Local mode only accepts requests with loopback Host/Origin headers.
```
pmeier, line 7: This seems unrelated. Please remove.

internal/api/cors_test.go
```diff
@@ -8,10 +8,10 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
-func corsTestRouter(localMode bool) *gin.Engine {
+func corsTestRouter(localMode bool, allowedOrigins ...string) *gin.Engine {
```
pmeier, line 11: Unrelated, please remove.

internal/api/router.go
```diff
@@ -63,7 +63,7 @@ func NewRouter(cfg *config.Config, db *gorm.DB, q queue.Queue, exec executor.Exe
 	// Middleware
 	router.Use(gin.Recovery())
 	router.Use(loggingMiddleware())
-	router.Use(corsMiddleware(localMode))
+	router.Use(corsMiddleware(localMode, cfg.Server.AllowedOriginsList()))
```
pmeier, line 66: It seems two PRs got conflated here. Not saying CORS doesn't need fixing, but it shouldn't be part of this PR. Not going to flag this again as part of this review.

internal/db/db.go
```diff
@@ -144,7 +150,10 @@ func seedDefaultRoles(db *gorm.DB) error {
 		var existing models.Role
 		result := db.Where("name = ?", role.Name).First(&existing)
 		if result.Error == gorm.ErrRecordNotFound {
-			if err := db.Create(&role).Error; err != nil {
+			// Role.Name has a unique index, so a concurrent booting process
```
pmeier: Is this related? We aren't touching roles here, are we?

internal/db/db_test.go
```diff
@@ -0,0 +1,96 @@
+package db
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/glebarez/sqlite"
```
pmeier: What is this used for? I can't find it. In any case shouldn't we use GORM for everything?

pmeier (changes requested):
Just file review. Will post another one after testing in a live deployment.

.gitignore
tylerpotts: Removed in 4fb732a. (Moved it to my local .git/info/exclude instead so it stays out of the repo config.)

config.yaml.example
tylerpotts: This was part of the CORS cherry-pick. Now that https://github.com/nebari-dev/nebi/pull/490 is merged I merged main into this branch, and the PR no longer touches this section.

internal/api/cors_test.go
tylerpotts: Gone after merging main. This file no longer differs from main now that https://github.com/nebari-dev/nebi/pull/490 landed.

internal/api/router.go
tylerpotts: Agreed. The CORS change was cherry-picked here while the standalone fix was still in review, since live testing of this feature needed it. https://github.com/nebari-dev/nebi/pull/490 has merged, so after merging main this PR no longer touches router.go or any netguard/CORS code.

internal/db/db.go
tylerpotts: Correct, roles are untouched by this feature. That was drive-by hardening for a concurrent-boot race that predates this PR; reverted in 4fb732a to keep the PR focused.

internal/db/db_test.go
tylerpotts: It was the GORM sqlite dialector (github.com/glebarez/sqlite is the pure-Go GORM driver db.go itself uses) passed to gorm.Open below. But fair point: in 4fb732a testDB now calls this package's own New(), so the direct driver import is gone and tests run through the production connection path.

pmeier (approved):

Issue #451, opened 2026-07-14
[security] Build/release pin an unsupported Go 1.24 line with no vulnerability scan gate


dcmcand:
## Summary

The module and every build surface pin the Go 1.24 release line, which is outside upstream security support, and CI has no reachability-aware vulnerability gate against source or final artifacts. (CWE-1104 Use of Unmaintained Third-Party Components.)

## Where it is (verified against `main` @ `15f66d4`)

- `go.mod:3` declares `go 1.24.0`.
- All build surfaces pin Go 1.24: `.github/workflows/release.yml:40`, `desktop.yml:63`, `ci.yml:109`, `ci.yml:129`, `ci.yml:163`, and `Dockerfile:11` (`golang:1.24-alpine`).
- Go's support policy covers a major release until two newer majors exist. With Go 1.25 and Go 1.26 released, the 1.24 line no longer receives fixes.
- No vulnerability gate exists: a grep across `.github/workflows/`, `.goreleaser.yaml`, and `Dockerfile` for `govulncheck|trivy|grype|syft|sbom|attest|cosign|scan` finds nothing (only the unused `id-token: write`). CI runs `golangci-lint` (`ci.yml:114-117`) and `go test -race` (`ci.yml:143`), neither of which is a vulnerability gate. `Dockerfile:6` uses `npm ci --no-audit`.

## Impact

- Security fixes issued only to supported Go lines are unavailable to the release.
- Compiler, runtime, and standard-library risks are not evaluated against the actual final binaries.
- A privileged desktop/server release can lag security fixes with no automated failure.

## Remediation

- Move to a currently supported Go release and its newest security patch; pin the toolchain by exact version and digest.
- Run `govulncheck` against source and compiled release binaries; scan final container images and desktop bundles.
- Record compiler identity in provenance/SBOMs; define maximum patch latency and time-bounded exceptions.

## Acceptance criteria

- CI rejects unsupported Go lines.
- Final-artifact provenance records the exact compiler.
- Reachable High/Critical vulnerabilities fail the release.
- Exceptions identify reachability, compensating controls, owner, and expiry.

Related: #211 (CVE reporting and SBOM generation).

_Identified in a security assessment of v0.13 (commit `be3187c`), re-validated against `main` @ `15f66d4`. Confidence: High for version support status; individual CVE reachability unverified._

PR #478, opened 2026-07-30
Harden release artifacts, installer verification, and image pinning

### LINKED ISSUE #447: [security] Release artifacts use mutable build inputs and lack signatures/provenance/SBOMs; installers skip checksum verification
## Summary

Release-critical build steps still use mutable tool and base-image inputs, and distributed artifacts lack independently verifiable signatures, provenance, and SBOMs. The official installers download and install without verifying the published checksum or any signature. A checksum generated by the same workflow that builds the artifact only proves consistency with that workflow's output. (CWE-494 Download of Code Without Integrity Check.)

Recent work already covers **macOS**: the CLI is notarized via GoReleaser/`quill` (`.goreleaser.yaml:56-69`) and the desktop app is codesigned/notarized/stapled (`.github/workflows/desktop.yml:101-150`). Wails is now pinned (`desktop.yml:93`, `v2.12.0`). The following remain:

- Mutable inputs: `swag@latest` (`release.yml:52`, `ci.yml:175`, `Dockerfile:19`); GoReleaser `version: latest` (`release.yml:61`, `ci.yml:184`); `ghcr.io/prefix-dev/pixi:latest` (`Dockerfile:38`); tag-only bases `node:20-alpine` (`Dockerfile:3`) and `golang:1.24-alpine` (`Dockerfile:11`); `npm install` (not `npm ci`) in `.goreleaser.yaml:7` and `desktop.yml:84`.
- No independent authenticity for non-macOS artifacts: no `signs:`/`sboms:`/`docker_signs:` in `.goreleaser.yaml`; no container-image signing in `docker.yml`; no SBOM anywhere; no SLSA/attestation step. `id-token: write` is declared (`release.yml:10`) but never consumed by any attestation/cosign step.
- No Windows Authenticode signing (`desktop.yml:252-262` just uploads the raw `.exe`).
- Installers verify nothing: `install.sh` downloads + extracts + installs with no `checksums.txt` fetch or hash/signature check; `install.ps1` likewise (`Expand-Archive` then copy).

## Impact

Compromise or mutation of a tool tag, package, base image, runner, release token, or registry tag can produce a malicious artifact that passes the current release process and installs without any integrity gate. This is high impact because the desktop artifact runs with developer-user privileges and the server image receives database, queue, registry, and workspace access.

## Remediation

- Pin every tool and container base to an exact version and immutable digest; use `npm ci` in every release path.
- Sign binaries and container images with a verifiable identity; generate GitHub artifact attestations / SLSA provenance; publish SPDX or CycloneDX SBOMs per artifact and image digest.
- Add Windows Authenticode signing.
- Make installers verify checksum and signature/attestation before extraction.
- Require signing/notarization secrets for release jobs and fail rather than silently publishing unsigned artifacts.

## Acceptance criteria

- Every distributed artifact has a verifiable signature and provenance statement naming the source commit.
- The installer refuses a modified archive and a valid archive with a missing/invalid signature.
- Every deployed image reference is a digest, not `latest` or a mutable tag.
- Release publication fails when signing, attestation, SBOM, or vulnerability gates fail.

Related: #211 (CVE reporting and SBOM generation); #426 / #249 and #427 (macOS signing, partially addresses this).

MUFFANUJ:
## Reference Issues or PRs

Closes #447 

- [x] Documentation Update
- [x] Build related changes
- [x] Other (please describe): Release supply-chain security hardening

## Testing

- [ ] Did you add new tests?

#### Text styling

#### Non-text content

## Any other comments?

This PR addresses the release supply-chain hardening work from [#447](https://github.com/nebari-dev/nebi/issues/447) by pinning mutable build inputs, centralizing tool versions in `.github/tool-versions.env`, switching release paths to deterministic installs with `npm ci`, pinning Docker base/deployment image references to immutable digests, and adding release-time signing, SBOM, provenance, and vulnerability gates.

The GoReleaser CLI release now signs artifacts with Sigstore, publishes archive SBOMs, emits checksums, creates GitHub attestations, and keeps releases in draft until all required CLI, desktop, and container assets are present and current for the release commit. The shared `publish-release-if-ready.sh` gate verifies required assets, verifies signed marker/checksum bundles, checks marker `source_commit` and `run_id`, rejects stale workflow runs, and enforces CLI archive digests against the signed `checksums.txt` before publishing the draft release.

The desktop and container release workflows now produce signed release artifacts and metadata, generate SBOMs, create attestations, and upload signed current-run marker files so stale assets from previous failed runs cannot accidentally publish a release. The release gate also validates marker payload digests for desktop assets, SBOMs, container image metadata, and CLI checksum metadata.

The installers now verify release checksums and Sigstore signatures before extraction or installation, so modified archives and missing or invalid signature bundles are rejected instead of installed. Installer CI now makes legacy-release smoke-test skips visible with GitHub Actions warnings and exercises local failure paths that do not depend on a published signed release. Documentation and examples were updated to note the `cosign` requirement and digest-pinned image expectations.

One remaining follow-up is the broader workflow permission split tracked in [#512](https://github.com/nebari-dev/nebi/issues/512). This PR avoids PR signing/attestation noise where practical, but the cleaner long-term shape is to separate CI and release permission boundaries, either by splitting jobs/workflows or by moving shared build logic into reusable components while keeping tag-only signing and publishing isolated.


aktech:
The vuln scan is failing in the CI.

.github/workflows/docker.yml
```diff
@@ -5,9 +5,17 @@ on:
...
+
+permissions:
+  actions: read
+  contents: write
+  id-token: write
+  attestations: write
 
 concurrency:
-  group: ${{ github.ref }}
+  group: release-${{ github.ref }}
```
tylerpotts: **Blocker:** with this change, release.yml, desktop.yml, and docker.yml all share the `release-${{ github.ref }}` concurrency group, and all three trigger on a `v*` tag push. GitHub allows at most one running plus one pending run per concurrency group; when a third run queues, the previously *pending* run is canceled. `cancel-in-progress: false` only protects the running run, not the pending one. So on every tag push, one of the three workflows gets nondeterministically canceled, its gated assets are never uploaded, and `publish-release-if-ready.sh` never sees the complete asset set: the release stays in draft until someone notices and manually re-runs the canceled workflow. (The desktop.yml comment "a queued run waits rather than being dropped" was true with two workflows and is no longer true with three.)

Suggested fix: give each workflow its own group (e.g. `release-${{ github.workflow }}-${{ github.ref }}`) so nothing gets canceled, and serialize release writes another way. Since separate groups reintroduce the concurrent-draft-creation race the shared group originally solved, the cleanest shape is probably the build/publish split you already floated in the PR description: build jobs upload artifacts, and a single downstream gate job serializes the release writes and publish.

.github/scripts/publish-release-if-ready.sh
```diff
@@ -0,0 +1,161 @@
...
+  nebi-container-image.json
+)
+
+is_draft="$(gh release view "${GITHUB_REF_NAME}" --json isDraft --jq '.isDraft')"
+if [ "$is_draft" != "true" ]; then
+  echo "Release ${GITHUB_REF_NAME} is already public."
+  exit 0
+fi
+
+mapfile -t assets < <(gh release view "${GITHUB_REF_NAME}" --json assets --jq '.assets[].name')
```
tylerpotts: **Blocker:** `mapfile` requires bash >= 4, but this script is invoked as `bash .github/scripts/publish-release-if-ready.sh` from the desktop matrix, and on `macos-latest` that resolves to Apple's `/bin/bash` 3.2.57 (the GitHub macOS images ship no newer bash). With `set -euo pipefail` the script dies right here with `mapfile: command not found`, so the macOS desktop job fails at the publish gate on every tagged release. And if the macOS invocation happens to be the one that would have seen the complete asset set, publication is left to whichever other workflow's gate runs later, making publishing ordering-dependent.

Suggested fix, portable replacement:

```bash
assets=()
while IFS= read -r name; do
  assets+=("$name")
done < <(gh release view "${GITHUB_REF_NAME}" --json assets --jq '.assets[].name')
```

Alternatively, run the gate once from a single ubuntu job instead of five times across three workflows, which would also shrink the concurrency-group problem flagged on docker.yml.

tylerpotts (changes requested):
This is a strong hardening PR. The draft-until-complete release gate with signed `*.release.json` markers binding each asset to `source_commit`/`run_id`/`run_attempt` is a genuinely good design, the installer verification chain (signed `checksums.txt` -> archive hash -> per-asset Sigstore bundle, with the correct workflow identity per asset type) is layered correctly, and the fail-closed signing-secret validation covers the "fail rather than silently publish unsigned artifacts" criterion from https://github.com/nebari-dev/nebi/issues/447. The `main.commit` -> `main.Commit` ldflags change is also a real fix (released CLI binaries never embedded a commit hash before), and `install.sh` quietly fixes the Linux desktop binary name (`Nebi` vs `nebi-desktop`) that could never have worked.

All four acceptance criteria from https://github.com/nebari-dev/nebi/issues/447 are implemented, including Windows Authenticode signing. However, I found two release-mechanics blockers (inline comments) that I believe will strand or fail the pipeline on the first real tag, so requesting changes for those two specifically.

### Questions

- **PR-triggered OIDC/release permissions**: you flagged this yourself in the description. I would push to do the split now rather than as a follow-up: this PR is exactly what makes the workflow identity (`.github/workflows/desktop.yml@refs/tags/...`) worth stealing, and fixing blocker 1 likely wants the same build-job/publish-job split anyway. If it stays a follow-up, please open a tracking issue and link it from the workflow comments.
- **Old releases become uninstallable via the scripts**: `install.sh --version v0.13.x` now hard-fails because pre-hardening releases have no `checksums.txt.sigstore.json`. Fail-closed is right, but if installing older releases is a supported path the error should say "this version predates signed releases" instead of a generic download failure.
- **Vulnerability gates in the tag path**: `npm audit --audit-level=high` (which also scans devDependencies), Trivy HIGH/CRITICAL, and govulncheck can all block a release because of an advisory published between merge and tag, with no override. If that is a conscious choice, fine; `npm audit --omit=dev` would trim the flakiest source.
- **Compose digest defaults have no bump mechanism**: `docker-compose.prod.yml` now defaults `NEBI_IMAGE` to one frozen digest, and the traefik/postgres/valkey lines lost their human-readable version tags. Every release needs a manual digest bump with no checklist or automation added. Consider recording the source tag as a comment next to each digest, and adding the bump to the release process.
- **Gate run-matching edge**: docker.yml runs on both main pushes and tags, both `event=push` with the same `head_sha` when tagging a merge commit. If a main-push run is created after the tag run, the marker `run_id` will not match "latest" and the release silently stays draft. Filtering runs by `head_branch` equal to the tag name would make the match exact.
- **Candidate tags accumulate**: `publish-<sha>-<attempt>` tags are only deleted on failure; on success they pile up in Quay. Intentional audit trail, or should the success path clean up too?
- **Version-pin duplication**: ci.yml uses `GORELEASER_VERSION` while release.yml hardcodes `v2.17.1`; `SWAG_VERSION` exists in the Makefile while the Dockerfile hardcodes `swag@v1.16.6`; `COSIGN_VERSION` is declared independently in four workflows. First bump will drift (CI snapshot goreleaser diverging from release goreleaser is exactly the skew this PR exists to prevent). At minimum release.yml should use an env var like ci.yml, plus a comment cross-referencing the duplication sites.
- **`required_assets` duplicates the desktop matrix and fails open on additions**: removing a platform fails closed (visible), but adding one fails open, since the gate list will not include the new platform's assets and the release can publish before it uploads. Worth cross-referencing comments in both files.
- **Scope**: the Go 1.24 -> 1.25 bump, ~10 dependency upgrades, and the golangci-lint v1 -> v2 migration (which changes what CI enforces on all future PRs) are unmentioned in the PR body. If the linter migration was not required by the new gates, it would land more safely as its own PR; either way a sentence in the description would help.

### Stylistic

- `README.md`, `CONTRIBUTING.md`, and `docs/docs/installation.md` still say "Go 1.24+" while go.mod now requires 1.25; `AGENTS.md` still names golangci-lint v1.64.8.
- Both installers swallow cosign diagnostics (`>/dev/null 2>&1` / `*> $null`), so a failed install cannot distinguish clock skew from tampering from a truncated download. Let cosign stderr through on the failure path.
- The workflows build cosign/syft/crane from source with `go install` on every run; the SHA-pinned `sigstore/cosign-installer` action would match the repo's pinned-action convention and be much faster, unless building the tools from source is itself a provenance requirement.
- The release-gate contract (marker JSON schema) is produced by three near-identical heredocs but consumed by one script; a small `write-release-marker.sh` next to the gate script would put both halves of the contract in one place.
- `docker/pixi.Dockerfile` and `docker/uv.Dockerfile` still use floating tags (`ubuntu:24.04`, `python:3.12-slim`). Lower stakes since they are local dev/executor images, but either pin them or leave a comment stating they are deliberately excluded from the pinning policy.

tylerpotts:
Merged `main` in (`e40f895`) and worked through the review body's Questions and Stylistic sections in `3a74b0c`. The two inline blockers were already fixed in `90f8e16`; I verified both (the concurrency groups are now per-workflow, and `mapfile` is gone, confirmed under Apple's bash 3.2.57).

### Conflict resolution notes

Worth flagging one thing the merge caught: `aquasecurity/trivy-action` was pinned to `a9c7b0f`, which is the **annotated tag object** for v0.36.0, not a commit. Actions cannot resolve a tag object as a ref, so that step would have failed on every run. Repinned to `ed142fd`, the commit the tag dereferences to.

Otherwise: `main`'s Go 1.26.6 toolchain pin, digest-pinned `golang` base image, and `go-version-file: go.mod` win over the branch's hardcoded `'1.25'`; the branch's pinned `go install` for golangci-lint wins over `main`'s `curl | sh` from master (same version either way); the branch's candidate-push/sign/promote Docker flow supersedes `main`'s plain push but picks up `main`'s conditional gha cache and `#451` scan rationale.

### Questions

- **PR-triggered OIDC/release permissions** — kept as a follow-up, tracked in https://github.com/nebari-dev/nebi/issues/512 and linked from comments above the `permissions` blocks in `desktop.yml` and `docker.yml`. The issue spells out the build/publish split, including the two complications that make it non-trivial here (platform-bound macOS notarization and Windows Authenticode need a matrix on the release job, and `docker.yml` signs by digest in the same job that pushes the candidate).
- **Old releases uninstallable** — still fail-closed, but the error now names the reason: "nebi v0.13 has no checksums.txt.sigstore.json, so it predates signed releases (v0.14 and earlier)". Both installers.
- **Vulnerability gates** — added `--omit=dev` to both `npm audit` calls. Kept the gates otherwise fail-closed with no override; `RELEASING.md` says explicitly that the fix is to land the bump on `main` and re-tag.
- **Compose digest bump mechanism** — each third-party digest now carries its source tag in a comment. `postgres` is repinned to `18.6`: the previous digest matched no current tag, so no truthful comment was possible. Added the bump as a checklist in the new `RELEASING.md`, linked from `CONTRIBUTING.md`.
- **Gate run-matching edge** — good catch, this was real. `publish-release-if-ready.sh` now filters candidate runs by `head_branch == GITHUB_REF_NAME`, which is what `wait-for-release-workflow.sh` was already doing.
- **Candidate tags accumulate** — now deleted on success as well as failure. Deleting the tag does not delete the manifest: the final tags reference the same digest, and the signature and attestations are bound to the digest, not to any tag.
- **Version-pin duplication** — `release.yml` now uses `GORELEASER_VERSION`; the Dockerfile takes `SWAG_VERSION` as a build arg; every env block cross-references its duplication sites. One extra find while doing this: `ci.yml`'s swagger drift check used `swag@latest` with a comment saying "to match release.yml", which stopped being true once this PR pinned `release.yml` to `SWAG_VERSION`. It was validating output the release would never produce. Now pinned; verified `internal/swagger` is unchanged under v1.16.6.
- **`required_assets` fails open** — cross-referencing comments in both files, stating that removing a platform fails closed but adding one fails open.
- **Scope** — the Go bump, dep upgrades, and golangci-lint v2 migration have all landed on `main` independently, so they are no longer in this diff (`go.mod` and `.golangci.yml` show zero delta vs `main` now). Added a Scope section to the PR description covering what remains.

### Stylistic

- Go 1.24+ to Go 1.25+ in `README.md`, `CONTRIBUTING.md`, `docs/docs/installation.md`; `AGENTS.md` no longer names a golangci-lint version.
- Both installers surface cosign's stderr on the failure path. `install.ps1` redirects to files rather than using `2>&1` on the pipeline: under Windows PowerShell 5.1 with `$ErrorActionPreference = "Stop"`, native stderr becomes a terminating `NativeCommandError` before the handler can print it.
- cosign now comes from SHA-pinned `sigstore/cosign-installer@6f9f177` (v4.1.2) in all four workflows. This drops the Go toolchain setup from `test-install.yml` entirely, since cosign was the only thing it needed Go for. syft/crane/govulncheck have no equivalent pinned action, so they stay as version-pinned `go install`.
- The three marker heredocs are now `.github/scripts/write-release-marker.sh`, so the marker schema sits next to its consumer. Exercised under bash 3.2.57 with a stubbed cosign.
- `docker/pixi.Dockerfile` and `docker/uv.Dockerfile` are digest-pinned rather than excluded, with a note that the pixi/uv installs inside them still resolve latest.

Locally verified: `go build ./...`, `go vet ./...`, `golangci-lint run ./...` (0 issues), `shellcheck` clean on all `.github/scripts/*.sh` and `install.sh`, YAML parses on every workflow, and `bash -n` under 3.2.57 on every shell script. `install.ps1` was not executed (no pwsh available here), so the PowerShell changes are unverified beyond review.

tylerpotts:
Note that Claude accidentally pushed some code that I've since reverted. Will post a review of the current state of the code shortly

tylerpotts (changes requested):
**Disclosure:** I pushed the merge commit `e40f895` to this branch to clear the conflicts with `main`. Flagging it since I'm also the reviewer: the merge resolution itself deserves a second pair of eyes, particularly the `docker.yml` reconciliation where main's plain-push flow and this branch's candidate/sign/promote flow both changed the same region.

## What's fixed since my last review

Both inline blockers are genuinely resolved, and I checked the fixes rather than taking the resolution at face value:

- Concurrency groups are per-workflow now, with ordering re-established through the new `wait-for-release-workflow.sh`. That's a better answer than what I suggested.
- `mapfile` is gone. The whole script parses under Apple's bash 3.2.57.

Also still good, and worth restating: the per-asset cosign identity in the installers (`release.yml` for CLI, `desktop.yml` for desktop) is correct, `-X main.Commit` matches `var Commit` at `cmd/nebi/main.go:15`, dropping the dead `-X main.date` was right, and `.env.prod.example` quietly fixes `NEBI_IMAGE` pointing at a personal namespace (`quay.io/aktech/nebi`). That last one deserves a line in the PR body.

Every item from my previous review's Questions and Stylistic sections is still open. I'm not re-listing them; they stand as written.

## Blockers

**1. "Re-run failed jobs" permanently bricks the release.** `GITHUB_RUN_ATTEMPT` is run-level, shared across all three desktop matrix legs. Re-running one failed leg re-uploads only that marker at `run_attempt: 2`; the other two still say `1`. `publish-release-if-ready.sh:153-159` compares every marker against the run's current attempt, so two mismatch and the script `SystemExit`s. The publish step goes red on a re-run that succeeded, and the release can never leave draft. Only "Re-run **all** jobs" escapes, and the error message doesn't say so. This is the first thing anyone does after a partial failure.

**2. The gate has no ref filter, so a main-branch run can shadow the tag run.** `publish-release-if-ready.sh:85,111` match on workflow name plus `head_sha` only. Both `docker.yml` and `desktop.yml` trigger on main pushes *and* tags, so one commit routinely carries two `push` runs per workflow. `git push --follow-tags` delivers both within the same second and the `created_at` tiebreak becomes a coin flip. Your sibling script already gets this right at `wait-for-release-workflow.sh:45`. Apply the same guard, or better, factor run-selection into one helper. Two divergent copies of "find the latest run for this commit" is the root cause.

**3. `govulncheck` is pinned backwards and two versions clobber each other in one job.** `GOVULNCHECK_VERSION: v1.1.4` (new here) sits alongside `v1.6.0` hardcoded in four places on `main`. In `desktop.yml`, line 140 installs v1.1.4, line 144 scans source, line 163 reinstalls v1.6.0 over it, line 165 scans the binary, all in the same job. So macOS and Windows legs scan at v1.1.4 and Linux at v1.6.0. In `release.yml`, `vuln-gate` scans at v1.6.0 and `goreleaser` runs the identical command at v1.1.4. A hardening PR pinning the scanner *older* than what CI already runs inverts its own goal. This only became visible after the merge.

**4. `docker.yml` gains OIDC and release-write on an unfiltered `pull_request` trigger.** `main` has no `permissions:` block here at all; this adds `contents: write`, `id-token: write`, `attestations: write` at workflow scope with no `branches:` filter. Fork PRs get a forced read-only token, but same-repo branch PRs don't, and `steps.registry.outputs.publish` evaluates `true` for them, so PR builds push to Quay, `cosign sign` with the repo's OIDC identity, and create registry attestations. You raised this yourself as a follow-up question. I accepted that last time and I don't now, because of blocker 5. Reversing my own position here, so: the reason it changed is that the unverified markers make the permission grant load-bearing in a way I hadn't connected.

**5. The gate requires signed marker bundles and never verifies a signature.** `publish-release-if-ready.sh` lists eight `.sigstore.json` bundles in `required_assets`; the script contains zero occurrences of `cosign`. It downloads the raw marker JSON and parses it. The bundles are presence-checked and nothing more. On its own that's an overstated guarantee; combined with blocker 4 it means "an attacker cannot forge a marker" is not a property this design holds. Either verify each marker before trusting `source_commit`/`run_id`, or drop the bundles from `required_assets` so the gate doesn't imply more than it does.

**6. `ci.yml:124` still uses `swag@latest`, the one site the pinning missed.** The comment above it says "Uses swag@latest to match release.yml", which stopped being true when this PR pinned `release.yml` to `${SWAG_VERSION}`. The moment swag ships a codegen change, this regenerates `internal/swagger` differently from the committed v1.16.6 output and **every PR fails the drift check**.

## Questions

- **`mode: replace` plus `use_existing_draft` on a Release re-run.** `.goreleaser.yaml` sets `draft: true`, `use_existing_draft: true`, `replace_existing_artifacts: true`, `mode: replace`. GoReleaser's `replace` deletes existing release artifacts before uploading. If Release is re-run after desktop/docker uploaded, it may wipe their assets, and nothing re-triggers those workflows, so the release wedges permanently. I could not execute GoReleaser to confirm. **Please verify before merge**; the failure mode is unrecoverable without manual intervention.
- **What signals a wedged release?** The gate `exit 0`s on missing assets, so every invocation is green. Drafts are excluded from the `releases/latest` API, so `install.sh` silently keeps serving the previous version. Green badges, no alarm, stale installs.
- **Should `required_assets` cover the CLI archives?** It requires `checksums.txt`, the markers, and all 21 desktop assets, but never `nebi_<ver>_<os>_<arch>.tar.gz` or its `.sigstore.json`, which is exactly what `install.sh:251` hard-requires. A release could pass a complete-looking gate and still fail every install.
- **Marker payloads are write-only.** `checksums_sha256`, `artifact_sha256`, `sbom_sha256`, `digest` are computed, signed, and never read.
- **Is 2h the intended polling ceiling?** `240 x 30s`, with three desktop legs plus docker each holding a runner idle, two of them paid macOS/Windows.
- **Staleness detection for the new pins.** `.github/dependabot.yml` covers npm only. This adds roughly 12 pins across compose, Dockerfiles, and workflow env. The Go toolchain got a bespoke checker one commit before this branch's merge base; these get nothing.

## Stylistic

- `README.md:177` and `docs/docs/installation.md:75` put a literal `@<tag-or-commit>` placeholder **inside a copy-paste code fence**. Right intent, broken command. Use a concrete tag. `README.md:180` also still says "Requires Go 1.24+" on a line this PR touched.
- Version constants fan out rather than centralize: `COSIGN_VERSION` in four workflows, `SYFT_VERSION` in three, `SWAG_VERSION` across `Makefile`/`ci.yml`/`release.yml` plus hardcoded at `Dockerfile:21`. Replacing `@latest` with N independently-drifting pins recreates the problem one cycle later.
- `test-install.yml:28,94` hardcode `go-version: '1.25'` while every other workflow moved to `go-version-file: 'go.mod'`, bypassing `check-go-toolchain.sh`.
- `docker.yml:269-272` deletes the candidate tag only on failure, so `publish-<sha>-<attempt>` accumulates forever. The step is also positioned before the tag-only steps, so a failure in signing/wait/upload leaves the candidate behind anyway, while a mid-loop failure in "Publish final image tags" deletes the candidate with real tags already pushed.
- `publish-release-if-ready.sh:49-52`: process substitution hides `gh release view`'s exit status, so a transient API failure yields an empty `assets` and a silently stuck draft. On the macOS legs, `"${assets[@]}"` on an empty array is an unbound-variable error under `set -u` in bash 3.2.
- `install.sh:128` and `install.ps1:61` suppress all cosign output. Failures are correctly detected; only the diagnostics are lost, which is precisely when the user needs them.
- The desktop `.sha256` sidecars are unsigned and absent from the signed `checksums.txt`, so `verify_checksum_sidecar` proves nothing the following `verify_asset_signature` doesn't. Nothing is weakened, but three of them are load-bearing `required_assets` entries, which reads as if they carry weight.
- Compose files use bare `image@sha256:` with no tag. `Dockerfile` gets this right (`golang:1.26.6-alpine@sha256:...`). As written, dev and prod run different traefik versions and there's no way to tell from the files.
- The `actions/attest` pin is a real commit, but the `# v4` comment (6 sites) is imprecise: `v4` now points elsewhere. Every other pin names an exact patch tag.
- `docker/pixi.Dockerfile` and `docker/uv.Dockerfile` still float; `AGENTS.md:17` still names golangci-lint v1.64.8.

## Suggested follow-up

`AGENTS.md` documents the Go and frontend architecture thoroughly and the release pipeline not at all. After this lands, the pipeline is the most intricate subsystem in the repo, and every rule governing it is implicit: that renaming a workflow's `name:` breaks the marker contract while renaming the *file* breaks the installers' `--certificate-identity`, that the three tag workflows must never share a concurrency group, that the desktop matrix and `required_assets` are coupled with nothing enforcing it. Worth a short section, in this PR or immediately after.

---

The design is sound and I still think this is the right shape. Blockers 1 and 2 will strand a real release; 3 through 6 are each cheap to fix.

Six blockers on a second pass is a lot to absorb at once. If it helps, 1/2/3/6 are mechanical and could land first, with 4/5 handled as a separate conversation about what the gate is actually meant to guarantee.

tylerpotts (changes requested):
**Verdict: request changes** (3 blockers, all six from my last round resolved)

## What I verified, and how

Delta reviewed against my previous review at `e40f895`, current head `526e48f`.

Executed, not read:
- Mock `gh`/`cosign` harness running `publish-release-if-ready.sh` end to end through 6 scenarios, including mutation checks (tampered artifact, forced signature failure, stale `run_id`, shadow runs).
- The **old** gate script from `e40f895` against identical fixtures, to confirm each fix is a real behaviour change rather than a re-read.
- All three new shell scripts parsed under Apple's `bash 3.2.57`, plus `shellcheck` (clean).
- A container build against the real `.dockerignore`, confirming `.github/tool-versions.env` reaches the build context.
- `cosign` v2.6.1 `internal/ui/log.go` and `env.go` at the pinned tag; `actions/attest` `action.yml` at the pinned commit; PowerShell/PowerShell#4002.
- Live `releases/latest` asset list, and this PR's own CI job logs.

Could not run: Windows PowerShell 5.1 (no Windows host), GoReleaser, an actual tagged release.

## Previous blockers

| # | Item | Disposition |
|---|---|---|
| 1 | `run_attempt` bricks the release on "Re-run failed jobs" | **Fixed**, verified by mutation. Markers at attempts 1/2/1 with the API reporting attempt 2: old script exits 1, new script publishes. |
| 2 | No ref filter, main-branch run shadows the tag run | **Fixed**, verified by mutation. `select-release-workflow-run.py:13-15`, and both scripts now share one selector. A later `head_branch: main` run with the same `head_sha` no longer wins. |
| 3 | `govulncheck` pinned backwards, two versions in one job | **Fixed**. `.github/tool-versions.env:9`; zero hardcoded versions remain under `.github/`, `Makefile`, `Dockerfile`. |
| 4 | `docker.yml` OIDC + release-write on unfiltered `pull_request` | **Withdrawing the blocker.** See below. Tracked in https://github.com/nebari-dev/nebi/issues/512. |
| 5 | Gate requires signed bundles, never verifies one | **Fixed**, verified by mutation. `publish-release-if-ready.sh:70-81`, called at `:122-126` and `:211`. Forcing a cosign failure aborts at exit 1 before publish. |
| 6 | `ci.yml` `swag@latest` | **Fixed**. `:124` calls `make swagger`; `:159` and `:225` use `${SWAG_VERSION}`. |

**On blocker 4:** I escalated it last round specifically because blocker 5 made the permission grant load-bearing. With signature verification restored and identity pinned to `@refs/tags/<tag>`, a `refs/pull/N/merge` certificate cannot satisfy the gate, fork PRs get a read-only token with no OIDC, and same-repo authors already hold `contents: write`. My position moved twice because it was always conditional on blocker 5, and I should have stated it as a condition rather than a verdict.

**Also withdrawing** my round-1 and round-2 complaint about the Go 1.24 to 1.25 bump and the golangci-lint v1 to v2 migration being unmentioned scope. Both landed on main independently; `go.mod` and `.golangci.yml` are byte-identical to `origin/main` and are not in this diff. I was reading a stale merge base. The v2 config is also fine: `formatters: enable: [gofmt]` is present, so `gofmt` enforcement was not silently dropped.

## Previous questions

Fixed: `mode: keep-existing` (this was the one that could have wedged a release unrecoverably), CLI archives now in the gate plus digest-checked against signed `checksums.txt`, marker payloads now all read and enforced (tampering a desktop artifact after signing produces `sha256 mismatch`, exit 1), polling ceiling down to 45 minutes with the tradeoff documented, and version-pin duplication collapsed from five drift sources into one file consumed by `make`, Docker, and all five workflows.

Deferred acceptably: the OIDC split, with #512 tracking it.

Still open, not re-argued: the wedged-release signal (below), dependabot coverage, compose digest bump mechanism, `desktop_assets` duplicating the matrix, candidate-tag cleanup, the three marker heredocs, `AGENTS.md` having no pipeline section.

## Architecture

No design-gating issues. The gate is a real gate now, run selection lives in one place instead of two divergent copies (which was the root cause of blocker 2, so it was removed rather than patched twice), and pinning is centralised.

One non-gating concern: the coupling contract is still entirely implicit, and this PR made it larger. The gate keys markers off each workflow's `name:`; the installers key `--certificate-identity` off each workflow's *filename*. Renaming `desktop.yml` breaks every existing `install.sh --desktop`; renaming its `name:` breaks the gate. The desktop matrix and the gate's `desktop_assets` list must agree with nothing enforcing it. The three tag workflows must never share a concurrency group. None of that is written down. Highest-value follow-up, and it does not need to be in this PR.

## Blockers

### 1. The Windows installer aborts on a *successful* signature check under Windows PowerShell 5.1

`install.ps1:57-61`, with `$ErrorActionPreference = "Stop"` at `:15`.

Two facts, both primary-source:

1. **cosign prints its success message to stderr.** At the pinned `v2.6.1`, `verify_blob.go:145,286` call `ui.Infof(ctx, "Verified OK")`; `internal/ui/log.go:22` is `fmt.Fprintln(w.Stderr, msg)` with `env.go` defaulting `Stderr: os.Stderr`. There is always stderr output, including on the happy path.
2. **Under `$ErrorActionPreference = 'Stop'`, any `2>` redirection makes native-command stderr a script-terminating error.** PowerShell/PowerShell#4002: *"anything getting written to PowerShell's error stream instantly triggers a function/script-terminating error"*, and *"any `2>` redirection causes the symptom, such as `2>&1` or `2>err.txt`."* Its environment list includes Windows PowerShell 5.1. The fix, `PSNotApplyErrorActionToStderr`, became default in PowerShell 7.3.

So `Confirm-CosignBlob` throws before reaching the `$LASTEXITCODE` check, propagates through the outer `try`/`finally` (no `catch`), and the install dies with cosign's own success text rendered as an error. The documented invocation is `irm https://nebi.nebari.dev/install.ps1 | iex` (`README.md:165`, `docs/docs/installation.md:51`), which on a default Windows box is 5.1. CI runs `pwsh` 7.4+, which is immune, and skips this script entirely anyway (blocker 2).

I could not execute 5.1. Please confirm in `powershell.exe`:

```powershell
& { $ErrorActionPreference = 'Stop'; $o = cmd /c "echo hi 1>&2" 2>&1; 'reached' }
```

If it prints `reached`, I'm wrong. Fix if not:

```powershell
$prevEap = $ErrorActionPreference
$ErrorActionPreference = 'Continue'
try   { $output = & cosign verify-blob ... $ArtifactPath 2>&1 }
finally { $ErrorActionPreference = $prevEap }
```

Two things I owe you: this is not a delta regression, since `*> $null` redirects stderr too and hits the same mechanism. It is introduced by this PR (main has no cosign in `install.ps1`), and I missed it in both prior rounds while reviewing these exact lines.

### 2. The installer smoke test is now unconditionally skipped and reports pass

`test-install.yml:52-61` and `:122-126` require `checksums.txt` **and** `checksums.txt.sigstore.json` on `releases/latest`. `v0.14` has the former and not the latter. From this PR's CI logs at `526e48f`:

```
test-unix (ubuntu-latest): Latest release v0.14 predates signed installer verification assets; skipping installer smoke test.
test-windows:              Latest release v0.14 predates signed installer verification assets; skipping installer smoke test.
```

All three legs report **pass**. Before this PR, `test-install.yml` actually ran `install.sh` against the latest release and it worked. So the PR rewrites both installers (~200 lines, the whole verification chain) and in the same change converts their only test into a permanent no-op that never even goes yellow. That is exactly how blocker 1 survived two review rounds inside a green pipeline, and the skip cannot clear until a hardened release ships, which cannot happen until this merges.

Your alternative was leaving CI red, which is worse, so I'm not asking for full coverage. But "silently green" is the failure mode this PR exists to eliminate and shouldn't be the mechanism it ships. Minimum:

- Make the skip loud: `echo "::warning::installer smoke test skipped ..."` in both legs, so it's distinguishable from a real pass.
- Add at least one path that runs without a published signed release. Two that need no signatures: assert `install.sh` fails with the `cosign is required` message when cosign is off `PATH`, and assert it fails on a locally corrupted `checksums.txt`. Neither catches blocker 1, but both keep the scripts executing on every PR instead of never.

### 3. The Windows installer's cosign prerequisite is documented only in the Linux/macOS section

`docs/docs/installation.md:25-27` puts the note inside `### Linux & MacOS` (`:23`). The `### Windows (Powershell)` section at `:48-52` has none, and `install.ps1:52-53` hard-errors without cosign. A reader copying `:51` gets `Error: cosign is required to verify nebi release signatures.` with no warning on the page. `README.md:154-157` gets this right by sitting above both fences. One-line fix: promote the note to under `## Installation script` (`:21`).

## Suggestions

1. **The PR body now describes behaviour the code doesn't have.** It says the gate checks `source_commit`, `run_id`, *and* `run_attempt`. `run_attempt` is no longer checked, because removing it was the fix for blocker 1. The body also omits the three substantive changes of this round: marker signature verification, `.github/tool-versions.env`, and CLI-archive digest enforcement.

2. **`make lint` now diverges from CI, against upstream's advice.** `Makefile:26,112` moved from the official install script (the same pinned release binary `golangci-lint-action` fetches at `ci.yml:130-132`) to `go install`. Upstream: *"Using `go install`/`go get` ... aren't guaranteed to work. We recommend using binary installation,"* first reason being *"Local compilation depends on your Go version, which may differ from the tested build environment."* Keeping the install script with `-b $$(go env GOPATH)/bin $(GOLANGCI_LINT_VERSION)` gets the centralised pin without the divergence.

3. **The new ref filter has one bypass.** `select-release-workflow-run.py:14` reads `if expected_ref_name and head_branch and head_branch != expected_ref_name`. Executed: a candidate with `head_branch: null` skips the filter and still wins on `created_at`. Not reachable for real push events, but it's the single place the blocker-2 fix can fall open, and dropping the guard costs nothing.

4. **No `github-actions` ecosystem in dependabot.** `.github/dependabot.yml` is npm-only. This PR collapses eleven tool pins into one bumpable file, but ~15 SHA-pinned actions and those eleven versions still have no updater. Five lines covers the actions half.

5. **`CONTRIBUTING.md:46` still says "Go 1.24+"** while `go.mod` requires `go 1.25.0`. This PR fixed the same claim at `README.md:180` and `docs/docs/installation.md:72`.

6. **Residual from withdrawn blocker 4.** With workflow-scoped `permissions:` and an unfiltered `pull_request`, every same-repo PR build runs `cosign sign --yes` and writes a permanent public Rekor entry plus registry attestations for an unmerged image. No security consequence given the identity pinning, just immutable public noise per PR. Gating the sign/attest steps on `startsWith(github.ref, 'refs/tags/')`, or narrowing `permissions:` to the job, is two lines while #512 waits. A `# see #512` comment beside the block would also close out the other half of my round-1 ask.

## Question

**What signals a wedged release?** The gate `exit 0`s when assets are missing (`publish-release-if-ready.sh:108-112`), so every invocation from every workflow is green, and drafts are excluded from `releases/latest`, so `install.sh:191` keeps serving the previous version. Green badges, no alarm, stale installs. Every path that *fails* now fails loudly and correctly (verified: exit 1 on stale `run_id`, digest mismatch, signature failure, markers-present-but-assets-missing). The gap is only "markers never arrived", which is indistinguishable from success. A single `::notice::` naming the outstanding markers would make it readable in the run summary.

---

All six blockers from last round are genuinely resolved, five verified by running old and new side by side. The design is sound and I wouldn't push back on it again. What's left is on one axis: the installers are the most user-facing and least-tested part of this change, and the change that made them less testable is in this PR. Blockers 1 and 3 are one-liners. Blocker 2 is the one worth a conversation. If the 5.1 repro comes back the other way, say so and I'll withdraw blocker 1.

tylerpotts (approved):
**Verdict: approve** (all three blockers from my last round resolved, no new blockers)

## What I verified, and how

Delta reviewed against my previous review at `526e48f`, current head `089ce222`. The four non-merge commits since then are `8193b68`, `4f4babd`, `7813326`, `39d36de`; the merges from main were excluded.

Executed, not read:
- The new `install.sh` failure-path harness from `test-install.yml`, replicated locally under Apple's bash 3.2.57. Both cases (no cosign on PATH, corrupted `checksums.txt`) fail with the expected message and exit 1. `shellcheck` clean on `install.sh` and all three gate scripts.
- `select-release-workflow-run.py` at `526e48f` and at head against one fixture containing a tag run, a `main` run, and a `head_branch: null` run for the same SHA. The old selector returned the null-branch run; the new one returns the tag run.
- This PR's CI at `089ce222`: `Test Install Scripts` green on all three legs, both new failure-path steps ran (not skipped), and the `::warning::` skip annotation is present in each leg's log.
- Primary sources for the Windows fix: GitHub's shell mapping (`shell: powershell` is `powershell.exe` 5.1, confirmed by the shell path in the step-6 log header), PowerShell/PowerShell#4002, PR PowerShell/PowerShell#13361, and the 5.1 `about_Automatic_Variables` text on `$LASTEXITCODE` and the `exit` keyword.

Could not run: Windows PowerShell 5.1 itself, GoReleaser, an actual tagged release.

## Previous blockers

| # | Item | Disposition |
|---|---|---|
| 1 | `install.ps1` aborts on a *successful* cosign check under Windows PowerShell 5.1 | **Fixed.** `install.ps1:57-67` wraps the cosign call in an `ErrorActionPreference = "Continue"` try/finally, exactly the shape I suggested. Better: the new Windows failure-path step runs under `shell: powershell` (5.1) with a stub `cosign.cmd` that writes to stderr and exits 0, then asserts the *checksum* error, which is only reachable if the cosign step survived stderr output. That step is green at head, so the fix is exercised on every PR, not just asserted. One caveat: the wrap and the test landed in the same commit, so the test has never been observed red without the fix. The negative case rests on the documented 5.1 behaviour, not execution. |
| 2 | Installer smoke test unconditionally skipped and reports pass | **Fixed.** Skip now emits `::warning::` in both legs, and the new failure-path steps execute `install.sh` and `install.ps1` on every PR without needing a signed release. Verified locally (bash) and in CI (all three legs). |
| 3 | Windows cosign prerequisite documented only in the Linux/macOS section | **Fixed.** `docs/docs/installation.md:21-25`, note now sits under `## Installation script` above both OS sections. |

## Previous suggestions and question

All addressed: PR body rewritten and now matches the code (no `run_attempt` claim, mentions marker verification, `tool-versions.env`, CLI digest enforcement, and the CI changes); `make lint` uses the upstream install script with the centralised pin; the `head_branch: null` bypass is closed (verified by mutation above); `github-actions` added to dependabot; `CONTRIBUTING.md` says Go 1.25+; PR builds no longer sign or attest, with a comment pointing at https://github.com/nebari-dev/nebi/issues/512; and the gate emits a `::notice::` naming outstanding markers, which answers "what signals a wedged release".

**Correction to my last review:** I said the `PSNotApplyErrorActionToStderr` behaviour became default in PowerShell 7.3. The official docs place it in 7.2. Mechanism and conclusion unchanged.

## Architecture

No design-gating issues. Unchanged from last round: the release-pipeline coupling contract (gate keyed on workflow `name:`, installer identity keyed on workflow filename, desktop matrix vs `desktop_assets`, distinct concurrency groups) is still undocumented. Still the highest-value follow-up.

## Suggestions (non-gating)

1. **The Windows happy-path smoke test still runs under pwsh 7.** `test-install.yml:307,315` ("Run install script", "Verify installation") use `shell: pwsh`; only the new failure-path step uses `shell: powershell`. Once a signed release ships, the real cosign binary will be exercised under a shell immune to blocker 1's mechanism, while users default to 5.1. Switch those two steps to `shell: powershell`. This is the one I'd land before the first hardened tag.
2. **`Makefile:26,112` fetch the golangci-lint installer from `master`.** The binary version is pinned, the script is not. Main did this before the PR too, so not a regression, but `https://raw.githubusercontent.com/golangci/golangci-lint/$(GOLANGCI_LINT_VERSION)/install.sh` resolves and keeps everything on the one pin.
3. **Three copies of `github.event_name != 'pull_request'` in `docker.yml`.** "Determine registry publishing" already computes the publish decision once; a second output (`sign`) from that step would stop the next signing step from having to remember the condition.
4. **`install.ps1:69` still prints cosign diagnostics via `Write-Host`** while this round moved `Write-Err` to stderr. `install.sh` sends the same output to stderr.
5. **The two test harnesses (~90 lines each) are inline in `test-install.yml`.** Everything else of that size in this PR lives in `.github/scripts/`, where it can be run locally and linted. The error strings they match ("cosign is required", "Checksum verification failed") are now load-bearing; a comment at the installer call sites would keep a wording tweak from silently breaking CI.
6. **Still open from earlier rounds, not re-argued:** `AGENTS.md` pipeline section, the three marker heredocs, `desktop_assets` duplicating the desktop matrix, compose digests with no tag comment or bump mechanism, candidate-tag cleanup only on failure, and no updater for the pins in `tool-versions.env` or the compose digests.

---

Three rounds in, every blocker has been fixed properly rather than papered over, and the gate and installers are now tested in ways they weren't when this started. Approving. Suggestion 1 is worth a quick follow-up before the first tagged release.

PR #481, opened 2026-08-01
Fail closed on authorization reconciliation errors

### LINKED ISSUE #452: [security] Group and administrator reconciliation fails open on error
## Summary

After authentication, group-membership synchronization and proxy administrator reconciliation log their errors and then continue, so authentication success is treated as sufficient even when the authorization state could not be brought into agreement with the authoritative identity claims. (CWE-754 Improper Check for Unusual or Exceptional Conditions.)

- Group-sync errors are logged and login/request continues (fail-open):
  - `internal/auth/oidc.go:160-162` (`slog.Warn("OIDC group sync failed; continuing login")`) then proceeds to token generation (`:165`).
  - `internal/auth/basic.go:204-206` (Middleware) then `c.Next()` (`:212`).
  - `internal/auth/basic.go:250-252` (SessionFromProxy) then issues token.
  - `internal/auth/basic.go:304-306` (ExchangeIDToken) then issues token.
  - (`SyncOIDCGroups` in `internal/auth/group_sync.go` returns its errors; the fail-open is at the call sites.)
- Proxy admin reconciliation is void and logs-and-returns on failure, leaving the session usable: `internal/auth/proxy.go:130-134` (admin-check failure → bare `return`, so neither grant nor revoke runs), `:137-141` (grant failure logged), `:143-147` (revoke failure logged). Callers cannot detect the failure and proceed to set the context user / mint the JWT.

## Impact

During database, Casbin, or reconciliation failures: removed group permissions can remain effective; newly required restrictions may not apply; a removed administrator can remain administrator; sessions can be issued with stale authorization state; and operators receive only a warning rather than a failed security boundary.

## Remediation

- Make security-sensitive reconciliation transactional with session issuance and fail closed when the current authorization state cannot be proven.
- Where availability requires it, issue a deliberately restricted session with no administrative or registry access instead of a fully privileged one.
- Alert on every reconciliation failure; reconcile continuously and record last successful sync; distinguish IdP outage from local authorization-storage failure.

## Acceptance criteria

- A forced revoke failure does not result in an administrator session.
- A failed group sync cannot silently preserve access.
- Monitoring alerts before stale authorization exceeds a defined interval.
- Tests inject database/Casbin failures at every reconciliation step.

MUFFANUJ:
## Reference Issues or PRs

Closes #452 

- [x] Documentation Update
- [x] Code style update (formatting, renaming)
- [ ] Build related changes
- [x] Other (please describe): Security hardening

## Testing

- [x] Did you add new tests?

## Documentation

Updates the server setup documentation to describe the fail-closed reconciliation and stale authorization behavior.

#### Text styling

#### Non-text content

## Any other comments?

This PR fixes the security issue where OIDC group synchronization and proxy administrator reconciliation could fail open. Previously, Nebi logged reconciliation errors and continued issuing or accepting sessions, which could preserve stale group access or administrator privileges when the local database, Casbin, or reconciliation logic failed.

The fix makes authorization reconciliation part of the authentication boundary. OIDC callback login, proxy session exchange, proxy middleware request handling, and device-token exchange now reject the request if group synchronization fails. Proxy administrator reconciliation now returns errors instead of only logging them, so Nebi does not issue or accept an administrator session when admin check, grant, or revoke fails.

This also adds reconciled JWT metadata through `token_source` and `authorization_synced_at`. Reconciled OIDC/proxy/device-flow tokens are rejected once their authorization-sync timestamp becomes stale, and legacy unstamped reconciled tokens are rejected. Normal basic username/password tokens remain valid without an authorization-sync timestamp because they are not derived from external group/admin reconciliation.

To support monitoring and recovery, this adds an `AuthReconciliationStatus` model and migration. Nebi records the last reconciliation success, last failure, failure source, last error, and desired group/admin state needed for retry. A team-mode reconciliation monitor starts after router and RBAC initialization, retries unresolved local DB/Casbin reconciliation failures, and logs stale authorization alerts. IdP/token failures are distinguished from local authorization-storage failures.

The tests were expanded to inject DB and Casbin failures across group sync, admin sync, stale token handling, and monitor retry behavior.

As a follow-up improvement, the fixed monitoring interval could be made configurable through server config or environment variables, and to reduce noisy repeated alerts when retries continue failing, we could also add alert throttling, such as alerting at failure counts `1, 2, 4, 8, 16,` or alerting once every N minutes per `user_id` and reconciliation kind. For now, I think the current state of the PR is good enough; these are for future consideration.


internal/auth/reconciliation_status.go
```diff
@@ -0,0 +1,335 @@
...
+	"gorm.io/gorm"
+)
+
+type authReconciliationKind string
+
+const (
+	authReconciliationOIDCGroups authReconciliationKind = "oidc_groups"
+	authReconciliationProxyAdmin authReconciliationKind = "proxy_admin"
+
+	authReconciliationStaleAfter      = 5 * time.Minute
```
tylerpotts: **Blocker:** This constant gives every reconciled token a hard 5-minute lifetime regardless of reconciliation health. `requireFreshAuthorizationSync` compares only the timestamp baked into the JWT at mint time; it never consults the `AuthReconciliationStatus` table this PR adds, and nothing re-stamps a live token.

Concretely: device-flow tokens from `nebi login` are stored in the keyring and reused as plain bearers with no refresh path (`cliclient` has an `IsUnauthorized` helper with zero callers), so every CLI command fails with 401 five minutes after login. The frontend's 401 interceptor does `window.location.href = /login`, so proxy-mode web users get a full page reload and redirect dance every 5 minutes, losing in-flight UI state; non-proxy OIDC users must re-authenticate manually. Meanwhile `TokenDuration` is still 24h, so the advertised token lifetime is meaningless for these flows.

Suggested fixes, in rough preference order: (a) on validation of a stale reconciled token, consult the user's `AuthReconciliationStatus` and accept when a reconciliation success is fresher than `staleAfter` (the data is already persisted, just never read on this path); (b) add a token-refresh endpoint used by cliclient and the frontend; (c) align the staleness window with `TokenDuration` for bearer flows and rely on the per-request reconciliation that the proxy cookie path already does. In any case this window determines session lifetime and should be operator-configurable rather than a follow-up.

internal/auth/basic.go
```diff
@@ -150,12 +180,28 @@ func (a *BasicAuthenticator) validateToken(tokenString string) (*Claims, error)
...
+		if err := claims.requireFreshAuthorizationSync(time.Now().UTC()); err != nil {
+			return nil, err
+		}
 		return claims, nil
 	}
 
 	return nil, ErrUnauthorized
 }
 
+func (c *Claims) requireFreshAuthorizationSync(now time.Time) error {
```
tylerpotts: **Question:** Is a one-time global logout on upgrade intended? Every token issued before this PR, including basic username/password tokens, carries neither `token_source` nor `authorization_synced_at`, so all live sessions 401 the moment this deploys. That may be acceptable for a security fix, but the PR is marked non-breaking, and the new doc sentence in `docs/docs/server-setup.md` ("legacy unstamped tokens are rejected unless they were explicitly issued by basic username/password login") does not match the code: a legacy basic token has no `token_source` claim and is rejected like any other. If the global logout is intended, please say so in the docs and PR description; if not, the claim schema needs a grace/versioning strategy.

internal/auth/reconciliation_status.go
```diff
@@ -0,0 +1,335 @@
...
+	err := db.Where("user_id = ? AND kind = ?", userID, string(kind)).First(&status).Error
+	switch {
+	case err == nil:
+		status.LastSuccessAt = &now
+		status.LastFailureAt = nil
+		status.LastFailureSource = ""
+		status.ConsecutiveFailures = 0
+		status.LastError = ""
+		applyAuthReconciliationState(&status, state)
+		return db.Save(&status).Error
```
tylerpotts, line 99: **Blocker:** In proxy-cookie mode this success path runs on every request (`Middleware` calls `syncOIDCGroups` + `syncProxyAdminRole` per request), so each request now does a `First` plus `Save`/`Create` on `auth_reconciliation_statuses` twice, even when nothing changed. Beyond the write amplification, the First-then-Create pattern has no unique-violation tolerance: two concurrent first requests from a new user (a browser page load fires parallel API calls) both see `ErrRecordNotFound`, both `Create` against the composite PK, and the loser's constraint violation propagates out of `syncOIDCGroups` as a reconciliation failure, turning a healthy request into a 401 and recording a bogus failure row the monitor will then "retry".

Suggest an upsert (`clause.OnConflict`) and only writing the success row when state actually changed or after a minimum interval since the last write.

internal/auth/reconciliation_status.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+		applyAuthReconciliationState(&status, state)
+		return db.Create(&status).Error
+	default:
+		return err
+	}
+}
+
+func recordAuthReconciliationFailureWithGroups(db *gorm.DB, userID uuid.UUID, kind authReconciliationKind, cause error, claimGroups []string) {
+	recordAuthReconciliationFailureWithState(db, userID, kind, cause, authReconciliationFailureSourceLocal, newOIDCGroupReconciliationState(claimGroups))
```
tylerpotts, line 178: **Blocker:** Both persistence wrappers hardcode `authReconciliationFailureSourceLocal`, and `logIdentityProviderAuthFailure` only logs (it fires before a user is resolved, so it could not write a row anyway). That means no production code ever persists `identity_provider` as a failure source, so the monitor's skip-retry branch (`status.LastFailureSource != string(authReconciliationFailureSourceIdentityProvider)`) can never trigger outside tests, and the PR's stated goal of distinguishing IdP outage from local storage failure exists in logs but not in the persisted state the monitor acts on. Either wire real IdP failures into the table at a point where the user is known, or drop the persisted source column and the monitor branch to avoid shipping a distinction that looks load-bearing but is not.

internal/auth/reconciliation_status.go
```diff
@@ -0,0 +1,335 @@
...
+		if err := syncOIDCGroupsOnce(db, status.UserID, claimGroups, rbacProvider); err != nil {
+			recordAuthReconciliationFailureWithGroups(db, status.UserID, kind, err, claimGroups)
+			return false, err
+		}
+		if err := recordAuthReconciliationSuccessWithGroups(db, status.UserID, kind, claimGroups); err != nil {
+			recordAuthReconciliationFailureWithGroups(db, status.UserID, kind, err, claimGroups)
+			return false, err
+		}
+		return true, nil
+	case authReconciliationProxyAdmin:
```
tylerpotts, line 388: **Question:** The monitor replays `DesiredAdmin` recorded at the last failed attempt, which can grant admin from stale state. Scenario: user is admin per IdP at login, reconciliation fails (Casbin down), `desired_admin=true` is persisted; the IdP then removes the user's admin group; Casbin recovers; the monitor grants admin from the stale record. The user's own token has gone stale by then, but the Casbin admin bit stays set until their next login. For a fail-closed security PR, a component that grants privileges from cached state deserves explicit justification. Would it be safer for the monitor to only retry revocations (the fail-closed direction) and leave grants to real logins with fresh claims?

internal/auth/oidc.go
```diff
@@ -60,6 +70,7 @@ func NewOIDCAuthenticator(ctx context.Context, cfg OIDCConfig, db *gorm.DB, jwtS
 
 	provider, err := oidc.NewProvider(ctx, discoveryURL)
 	if err != nil {
+		logIdentityProviderAuthFailure(authReconciliationOIDCGroups, err)
```
tylerpotts: **Question:** A provider discovery failure at construction time is a configuration/availability problem, not an authorization-reconciliation event for any user. Emitting it as an `alert=true` reconciliation alert will misroute operator alerting. Intentional, or should this call be dropped from `NewOIDCAuthenticator`?

internal/auth/group_sync.go
```diff
@@ -25,65 +26,41 @@ import (
 // into them would create permanent untracked grants (phase-2 reconcile only
 // considers source=oidc memberships).
 func SyncOIDCGroups(db *gorm.DB, userID uuid.UUID, claimGroups []string) error {
```
tylerpotts: **Question:** After this PR the exported `SyncOIDCGroups` has no production callers; it survives only as a wrapper that constructs its own `rbac.NewDefaultProvider()`. If it is intended public API, its godoc should mention the new side effect (writing `auth_reconciliation_statuses` rows); otherwise consider dropping it.

internal/auth/basic.go
```diff
@@ -150,12 +180,28 @@ func (a *BasicAuthenticator) validateToken(tokenString string) (*Claims, error)
 	}
 
 	if claims, ok := token.Claims.(*Claims); ok && token.Valid {
+		if err := claims.requireFreshAuthorizationSync(time.Now().UTC()); err != nil {
```
tylerpotts: **Stylistic:** When this returns `ErrAuthorizationStale`, the middleware's generic `invalid or expired token` 401 makes a stale reconciled token indistinguishable from a garbage token. Surfacing the stale case distinctly (e.g. `session requires re-authentication`) would let the CLI print an actionable "run nebi login again" message instead of appearing broken.

internal/auth/reconciliation_status.go
```diff
@@ -0,0 +1,335 @@
...
+				"user_id", userID, "kind", kind, "error", err)
+		}
+	}
+
+	var staleFor time.Duration
+	if lastSuccessAt != nil {
+		staleFor = now.Sub(*lastSuccessAt)
+	}
+
+	slog.Error("Authorization reconciliation stale alert",
```
tylerpotts: **Stylistic:** This logs "Authorization reconciliation stale alert" on the very first failure, when nothing is stale yet and `stale_for` is `0s`, which will confuse operators triaging alerts. Consider naming this one a reconciliation-failure alert and reserving "stale" for the monitor path.

internal/auth/group_sync.go
```diff
@@ -98,16 +75,118 @@ func SyncOIDCGroups(db *gorm.DB, userID uuid.UUID, claimGroups []string) error {
 		if _, ok := desired[m.Group.Name]; ok {
 			continue
 		}
-		if err := db.Where("group_id = ? AND user_id = ?", m.GroupID, userID).Delete(&models.GroupMember{}).Error; err != nil {
-			return fmt.Errorf("delete stale membership: %w", err)
-		}
-		audit.LogAction(db, userID, audit.ActionRemoveGroupMember, fmt.Sprintf("group:%s", m.GroupID),
-			map[string]any{"origin": "oidc", "user_id": userID})
-		if err := rbac.RemoveUserFromGroup(userID, m.GroupID); err != nil {
+		if err := rbacProvider.RemoveUserFromGroup(userID, m.GroupID); err != nil {
```
tylerpotts: **Stylistic:** The stale-membership filter (`current` minus `desired`) is computed twice, once here for the Casbin removals and again inside the transaction for the row deletes. Computing the stale list once and iterating it in both places would keep the two phases from drifting apart.

That said, the ordering itself is right: Casbin removal before DB delete means any partial failure errs toward less access, and the test asserting the stale DB row is retained for retry shows that was deliberate. Nice.

internal/models/auth_reconciliation_status.go
```diff
@@ -0,0 +1,25 @@
...
+	UserID              uuid.UUID  `gorm:"type:text;primary_key" json:"user_id"`
+	Kind                string     `gorm:"type:text;primary_key" json:"kind"`
+	LastSuccessAt       *time.Time `gorm:"index" json:"last_success_at,omitempty"`
+	LastFailureAt       *time.Time `gorm:"index" json:"last_failure_at,omitempty"`
+	LastFailureSource   string     `gorm:"type:text;index" json:"last_failure_source,omitempty"`
+	ConsecutiveFailures int        `gorm:"not null;default:0" json:"consecutive_failures"`
+	LastError           string     `gorm:"type:text" json:"last_error,omitempty"`
+	DesiredGroupsJSON   string     `gorm:"type:text" json:"desired_groups_json,omitempty"`
+	DesiredAdmin        bool       `json:"desired_admin,omitempty"`
+	HasDesiredAdmin     bool       `json:"has_desired_admin,omitempty"`
```
tylerpotts: **Stylistic:** `DesiredAdmin *bool` would express optionality the same way the sibling fields in this struct already do (`LastSuccessAt *time.Time`) and remove the `HasDesiredAdmin` sentinel.

internal/auth/proxy_test.go
```diff
@@ -173,18 +183,247 @@ func TestParseAdminGroups(t *testing.T) {
...
+	t.Helper()
+	rawToken := testIDToken(t, username, groups)
+	req := httptest.NewRequest(http.MethodGet, "/", nil)
+	req.AddCookie(&http.Cookie{Name: "IdToken-nebi", Value: rawToken})
+	return req
+}
+
+func testIDToken(t *testing.T, username string, groups []string) string {
+	t.Helper()
+	token := jwt.NewWithClaims(jwt.SigningMethodNone, jwt.MapClaims{
```
tylerpotts, line 427: **Stylistic:** Worth a brief comment that `alg=none` is deliberate here because the paired test verifier skips signature checks; `UnsafeAllowNoneSignatureType` inside an auth package is the kind of line security scanners flag.

internal/auth/group_sync_test.go
```diff
@@ -179,3 +187,344 @@ func TestSyncOIDCGroups_RefusesToMergeIntoNativeGroup(t *testing.T) {
...
+		}); registerErr != nil {
+			t.Fatalf("register delete callback: %v", registerErr)
+		}
+	default:
+		t.Fatalf("unsupported callback op %q", op)
+	}
+	return name
+}
+
+func registerDBTableFailureCallback(t *testing.T, db *gorm.DB, op string, table string, err error) string {
```
tylerpotts, line 498: This failure-injection harness is excellent. Scoping GORM callbacks per table and op to hit every DB failure point, without mocking GORM itself, is clean and very reusable; it is what makes the fail-closed coverage in this PR convincing.

internal/auth/group_sync.go
```diff
@@ -98,16 +75,118 @@ func SyncOIDCGroups(db *gorm.DB, userID uuid.UUID, claimGroups []string) error {
...
+		currentGroupSet[groupID] = struct{}{}
+	}
+
+	addedGroupIDs := make([]uuid.UUID, 0, len(desiredGroupIDs))
+	for _, groupID := range desiredGroupIDs {
+		if _, ok := currentGroupSet[groupID]; ok {
+			continue
+		}
+		if err := rbacProvider.AddUserToGroup(userID, groupID); err != nil {
+			for _, addedGroupID := range addedGroupIDs {
```
tylerpotts, line 120: Good detail: rolling back partial Casbin adds on failure, and preserving the original reconciliation error when the status write also fails (covered by its own test), keeps the failure semantics honest.

tylerpotts (changes requested):
Thanks for taking on https://github.com/nebari-dev/nebi/issues/452 so thoroughly. The core fail-closed conversion is correct at all four call sites and the failure-injection test coverage is genuinely excellent. I verified the branch locally: build, vet, and the full internal/auth test suite pass.

I'm requesting changes for three reasons, detailed inline: (1) reconciled tokens hard-expire 5 minutes after issuance with no refresh path, which breaks CLI device-flow logins and OIDC web sessions even when reconciliation is healthy; (2) the per-request status writes on the proxy path introduce a first-request race and hot-path write amplification; (3) the persisted identity-provider failure source is dead code, so the monitor's skip-retry branch can never trigger in production. A fuller writeup, including a suggestion to split the PR so the security fix can land independently, is in the summary comment.

tylerpotts:
Thanks for the deep work on this one. The core of the PR, converting every reconciliation call site from log-and-continue to fail-closed, is correct, complete against the acceptance criteria in https://github.com/nebari-dev/nebi/issues/452, and backed by genuinely thorough failure-injection tests. I verified the branch locally: `go build`, `go vet`, and the full `internal/auth` suite pass at 5b93091.

A few observations that span files rather than lines:

**Issue resolution.** All four acceptance criteria are met: revoke failure no longer yields an admin session, failed group sync can no longer silently preserve access, the monitor alerts at half the staleness window, and DB/Casbin failures are injected at every reconciliation step. The inline blockers are about side effects of *how* staleness is enforced, not about the fail-closed fix itself.

**Consider splitting the PR.** The fail-closed conversion (returning errors and rejecting sessions) is a contained, obviously-correct security fix that could merge quickly. The staleness enforcement, JWT claim-schema change (`token_source`, `authorization_synced_at`), status table + migration, and background monitor are each separable features with their own open design questions (session lifetime and refresh strategy, monitor concurrency and ownership, operator configurability). Splitting would let the fix for the fail-open vulnerability land independently while those are worked out.

**Monitor concurrency.** `StartAuthReconciliationMonitor` institutionalizes a second writer to Casbin state that runs concurrently with request-path reconciliation. The global enforcer is a plain `casbin.NewEnforcer` (not a `SyncedEnforcer`), and the monitor replays persisted desired state that can be older than an in-flight login's claims, with last-writer-wins and no per-user serialization. Request paths already raced with each other before this PR, but adding a permanent background writer is the point where the design should state who owns authorization state and how conflicts resolve.

**Configurability.** `authReconciliationStaleAfter` and the monitor interval directly determine session lifetime and alert cadence, which is the class of setting this project routes through `config.Config`/env (like `Auth.ProxyAdminGroups`). The PR body defers this as a follow-up; given the session-lifetime blocker, it probably needs to be part of this change (or the staleness change should be split out, per above).

**Docs.** The new paragraph in `docs/docs/server-setup.md` describes a refresh mechanism ("must refresh before stale authorization can exceed the defined interval") that does not currently exist for bearer-token flows, and its statement about legacy basic tokens does not match the code (see inline comment). Worth revisiting once the staleness design settles.

Supply-chain screen came back clean: no new dependencies, and the new goroutine, model, and migration all match the PR's stated purpose.

internal/auth/reconciliation_status.go
MUFFANUJ: Agreed, the original 5-minute token-only freshness check made the reconciled JWT behave like it had a hard 5-minute lifetime, even though the JWT expiry is 24h and there is no refresh path for CLI/device-flow tokens.

I updated this so the stale window is configurable and defaults to the normal token duration. Also, when a reconciled token’s embedded authorization_synced_at is stale, validation consults AuthReconciliationStatus and accepts the token only if the user has a fresh successful reconciliation and no unresolved reconciliation failure. That keeps the fail-closed behavior without breaking healthy bearer-token sessions every 5 minutes.

internal/auth/basic.go
MUFFANUJ: For this PR I think the compatibility grace is the right tradeoff: otherwise every existing token, including basic-login tokens, is invalidated on deploy. Since TokenDuration is 24h, the grace only matters for tokens minted before the upgrade and naturally stops mattering after one token lifetime.

Longer term, we should remove the unstamped-token fallback in a follow-up once deployed versions have aged past the legacy token window. At that point all valid tokens should carry `token_source` and reconciled tokens should carry `authorization_synced_at`, so we can make the schema strict without causing a surprise global logout.

internal/auth/reconciliation_status.go
MUFFANUJ: Agreed. I changed the monitor so it only retries proxy-admin revocations from cached desired state. Cached admin grants now remain unresolved and require a fresh login/proxy/device-flow reconciliation with current IdP claims.

internal/auth/reconciliation_status.go
MUFFANUJ: Thanks for pointing this out. I switched first writes to an upsert on `(user_id, kind)` so parallel first requests do not produce duplicate-key failures, and I skip success writes when the status is already fresh and unchanged. We still write when state changes or when a prior failure needs clearing.

internal/auth/reconciliation_status.go
MUFFANUJ: The persisted `identity_provider` distinction was not actually reachable in production because IdP failures are only logged before a user-specific status row can be written. I removed the monitor’s persisted IdP skip behavior and now treat persisted rows as retryable local reconciliation failures; IdP failures remain log-only.

internal/auth/oidc.go
MUFFANUJ: Dropped!

internal/auth/group_sync.go
MUFFANUJ: Updated the godoc; it was kept public as it was already public before this PR, and removing it was not necessary to fix the security issue

MUFFANUJ: After looking at the call sites, I removed the public wrapper and made the tests call the injected private helper directly as this seemed the appropriate action here ( https://github.com/nebari-dev/nebi/pull/481/changes/e27bbdf0de9b8170faf45b9036477c018962b9bb )

internal/auth/basic.go
MUFFANUJ: Added a distinct 401 response for `ErrAuthorizationStale` so stale reconciliation is no longer reported as a generic invalid/expired token.

internal/auth/reconciliation_status.go
MUFFANUJ: I renamed the immediate failure log to “Authorization reconciliation failure alert” and left “stale alert” for the monitor path, where `stale_for` reflects an unresolved failure over time.

internal/auth/group_sync.go
MUFFANUJ: Added a single helper, and now both phases iterate the same computed stale list

internal/models/auth_reconciliation_status.go
MUFFANUJ: Thanks!, Changed to `*bool`

internal/auth/proxy_test.go
MUFFANUJ: Added a comment.

internal/auth/group_sync_test.go
MUFFANUJ: Thanks!

MUFFANUJ:
> **Consider splitting the PR.** The fail-closed conversion (returning errors and rejecting sessions) is a contained, obviously-correct security fix that could merge quickly. The staleness enforcement, JWT claim-schema change (`token_source`, `authorization_synced_at`), status table + migration, and background monitor are each separable features with their own open design questions (session lifetime and refresh strategy, monitor concurrency and ownership, operator configurability). Splitting would let the fix for the fail-open vulnerability land independently while those are worked out.
> 


Hi @tylerpotts Thanks for the detailed review, I addressed the blocker-level issues around the staleness window, hot-path status writes, persisted failure source, cached admin grants, provider-discovery alerting, docs, and legacy tokens. Given that, I’m inclined to keep this together for now so the fail-closed behavior and its status/monitoring support land coherently, but I’m happy to split if you still think the freshness/monitor pieces need separate design review.

internal/auth/reconciliation_status.go
```diff
@@ -0,0 +1,416 @@
...
+	switch kind {
+	case authReconciliationOIDCGroups:
+		if status.DesiredGroupsJSON == "" {
+			return false, nil
+		}
+		claimGroups, err := decodeOIDCGroupReconciliationState(status.DesiredGroupsJSON)
+		if err != nil {
+			return false, err
+		}
+		if err := syncOIDCGroupsOnce(db, status.UserID, claimGroups, rbacProvider); err != nil {
```
tylerpotts: **Blocker:** This replays cached group *additions*, which is the same stale-grant pattern the proxy-admin case now avoids. `syncOIDCGroupsOnce` reconciles to the full `DesiredGroupsJSON` recorded at the failed attempt, so the scenario from the admin thread applies verbatim: user logs in with groups [a, b], reconciliation fails and the claim set is cached; the IdP removes them from b during the outage; Casbin recovers; the monitor re-adds both memberships from the stale cache. Unlike a token, the cached claim set has no age bound, so a multi-day outage replays multi-day-old claims. `TestAlertUnresolvedAuthReconciliationsRetriesOIDCGroupState` currently asserts the add happens, so the behavior is locked in by the suite.

The interaction that makes this worse than a consistency nit: when the replay succeeds, `recordAuthReconciliationSuccessWithGroups` writes a fresh `last_success_at` and clears the failure row, which is exactly what `hasFreshAuthorizationReconciliationStatus` checks on the stale-token path. So one monitor tick both applies the stale group grants and re-validates the user's stale bearer tokens through the DB fallback. The stale membership isn't latent until their next login; it's immediately usable with an old token.

Suggested fix, mirroring the admin decision: on retry, run only the removal phase (Casbin removes plus DB deletes for memberships not in the cached set) and leave additions to a fresh login with current claims. Removal-only is fail-closed by construction. If a full-state replay is ever wanted, it needs an age bound on the cached claims, but I don't think this PR needs that. Either way the retry-success path should probably not refresh `last_success_at` unless the replayed state came from fresh claims, since that timestamp now gates token acceptance.

tylerpotts (changes requested):
Thanks for the thorough turnaround on the first round. I re-verified the branch at e27bbdf: build, vet, and the full `internal/auth` and `internal/config` suites pass locally, and I confirmed each thread from the first review against the code rather than just the replies. The staleness redesign is much better: the configurable window defaulting to the token lifetime plus the status-table fallback resolves the session-lifetime blocker cleanly, the upsert and skip-when-fresh logic resolves the hot-path write concerns, and the monitor no longer grants admin from cached state. The new tests cover all of it. I also confirm the docs now match the code, including the legacy-token grace behavior.

On splitting the PR: keeping it together is fine. The design questions I wanted settled are now settled in the PR itself.

I'm requesting changes for one remaining issue, detailed inline: the monitor still replays cached OIDC group *additions*, which is the same stale-grant pattern we just removed for admin, and it interacts with the new staleness fallback in a way that makes it more than a consistency nit.

Two minor notes, neither blocking, fine as follow-ups:

- `hasFreshAuthorizationReconciliationStatus` accepts a fresh success of *any* kind for the user (no `kind` filter). Today every flow updates both kinds together so this is harmless, but a one-line comment or a `kind` filter would keep it from becoming a surprise if a future flow syncs only one.
- `LastFailureSource` can now only ever hold `local_authorization_storage`, since IdP failures are log-only. Worth dropping the column in the same future cleanup that removes the legacy unstamped-token fallback.

MUFFANUJ:
Hey @tylerpotts, I have addressed the blocker and minor suggestion in the latest commit. Is this good to merge? :)

internal/auth/reconciliation_status.go
```diff
@@ -0,0 +1,442 @@
...
+			return false, nil
+		}
+		if *status.DesiredAdmin {
+			return false, nil
+		}
+		if err := syncAdminRoleToDesired(status.UserID, *status.DesiredAdmin, rbacProvider); err != nil {
+			recordAuthReconciliationFailureWithAdmin(db, status.UserID, kind, err, *status.DesiredAdmin)
+			return false, err
+		}
+		if err := recordAuthReconciliationSuccessWithAdmin(db, status.UserID, kind, *status.DesiredAdmin); err != nil {
```
tylerpotts: **Blocker:** This is the last place a monitor retry can mint token freshness from cached state. A successful admin-revoke retry calls `recordAuthReconciliationSuccessWithAdmin`, which stamps `LastSuccessAt = now`. Since `hasFreshAuthorizationReconciliationStatus` is deliberately user-level (per the comment you added in `basic.go`), that one stamp both clears the unresolved-failure gate and satisfies the fresh-success gate, so the user's stale bearer tokens are accepted again for another `staleAfter` window backed by group state that was never freshly reconciled.

Concretely: a login fails both syncs during a Casbin outage; the IdP drops the user from a group during the outage; Casbin recovers; one monitor tick clears the OIDC failure (removal-only, correct) and succeeds the admin revoke with a fresh stamp. Result: `unresolved=0`, `freshSuccesses=1`, stale tokens live again. That's the same interaction from my last review, surviving one branch over. And as before, the suite locks it in: `TestAlertUnresolvedAuthReconciliationsRetriesProxyAdminState` asserts `LastSuccessAt != nil`.

The fix mirrors what you did for groups: add a `clearAuthReconciliationFailureWithAdmin` wrapping `clearAuthReconciliationFailureWithState` with `newProxyAdminReconciliationState(*status.DesiredAdmin)`, call it here instead, and flip the assertion in the test. The revoke itself is fail-closed by construction, exactly like the group removals; only the timestamp side effect is wrong. To acknowledge the counterargument up front: unlike partial group removal, the admin revoke does fully apply that kind's desired state. But the stamp vouches user-level for group state it never verified, so under your own user-level-freshness comment, clear-not-record is the right semantic.

tylerpotts (changes requested):
Thanks for the quick follow-up. I verified the branch at 9befa561: build, vet, and the full `internal/auth` and `internal/config` suites pass locally. The removal-only conversion is exactly right: `syncOIDCGroupRemovalsOnly` does no Casbin adds, no membership inserts, and no group creation, and the new `clearAuthReconciliationFailureWithGroups` clears the failure row without touching `last_success_at`, which resolves the token-regating half of my last comment. The test flip is thorough too, especially the end-to-end `TestBasicAuthenticator_RejectsStaleReconciledBearerTokenAfterCachedOIDCGroupRetry`.

One remaining issue, detailed inline: the same fix needs to apply to the proxy_admin branch of `retryAuthReconciliationStatus`, which still stamps a fresh `last_success_at` from cached state. It's a few mechanical lines mirroring what you already did for groups, and then I'm happy to approve.

Two non-blocking observations, fine as follow-up issues rather than changes here:

- The status writers (`clearAuthReconciliationFailureWithState` and the `record*` functions) do an unlocked read-then-`Save` of every column, so a monitor-tick clear racing a request-time failure record can briefly wipe a just-recorded failure. This predates your commit and self-heals on the next tick; optimistic locking or column-scoped `Updates` would close it.
- The monitor retries from a `Find` snapshot, so a fresh login completing between the snapshot and the retry can have a just-granted group revoked by the removal-only pass. Fail-closed direction, availability blip only; re-loading the row inside `retryAuthReconciliationStatus` and skipping if the failure is already cleared would cover it.

tylerpotts (approved):
Verified at fe3d459: build, vet, and the full `internal/auth` and `internal/config` suites pass locally, and the delta since my last review is exactly the admin-retry fix plus clean merges from main. 429d59a6 resolves the remaining blocker the right way: the admin retry now clears the failure without stamping `last_success_at`, both branches of `retryAuthReconciliationStatus` honor the same "retries must not mint freshness" invariant, and the flipped test locks it in. The two race-condition observations from my last review stand as follow-up issues, not blockers. Thanks for the careful work across all three rounds, this is a solid fail-closed conversion. Approving.

MUFFANUJ:
Thanks for the amazing review cycle. Let's merge this. I do not have merge permission. Can you provide me with merge permission or merge it? Thanks again!

Issue #211, opened 2026-03-25
CVE reporting and SBOM generation


aktech:
## Context

There is a risk of vulnerabilities from PyPI packages pulled by pixi. Reporting on CVEs and generating SBOMs (Software Bill of Materials) is needed to help users and organizations assess the security of their environments.

## Tasks

- Add CVE scanning/reporting for environments
- Generate SBOMs for published environments

PR #491, opened 2026-08-04
security: move to supported Go line and gate builds on vulnerability scans

### LINKED ISSUE #451: [security] Build/release pin an unsupported Go 1.24 line with no vulnerability scan gate

tylerpotts:
Closes https://github.com/nebari-dev/nebi/issues/451

### Toolchain currency (CWE-1104)
- `go.mod`: language version `1.25`, toolchain pinned exactly to `go1.26.5` (newest patch of the newest supported line)
- All five workflow Go setups now use `go-version-file: 'go.mod'`, so the `toolchain` directive is the single source of truth for every build surface
- `Dockerfile` builder stage pinned to `golang:1.26.5-alpine` by exact version **and** digest

### Gates
- **`scripts/check-go-toolchain.sh`** — fetches `go.dev/dl?mode=json` (which lists only currently supported releases) and fails when the pinned toolchain is on an unsupported line; warns when a newer patch of the line exists. `--strict` also fails on an outdated patch.
- **CI (`ci.yml`)** — new `Backend / Security` job: support-line check + reachability-aware `govulncheck` source scan of `./cmd/... ./internal/...`; the `Build CLI Binaries` job additionally runs `govulncheck -mode=binary` against an unstripped build of `./cmd/nebi`, so the compiled artifact's dependency set (including the exact stdlib the toolchain shipped) is evaluated. (The shipped binaries themselves are stripped with `-s -w`, which loses the symbol table and degrades govulncheck to module-level reporting with unfixable false positives like GO-2026-5932 `x/crypto/openpgp` — a package nothing in the module graph imports.)
- **Desktop (`desktop.yml`)** — binary-mode `govulncheck` of an unstripped Linux build of the root wails package, covering the desktop stack the CLI binary never links.
- **Release (`release.yml`)** — new `vuln-gate` job (strict toolchain check + source scan) that `goreleaser` requires, so a tag cannot publish with an outdated toolchain or a reachable known vulnerability.
- **Docker (`docker.yml`)** — Trivy scan of the final image, failing on fixable High/Critical findings (`ignore-unfixed: true` so unfixable base-image CVEs don't perma-block).

### Dependency updates
Required to make the new gates pass — all had vulnerabilities reachable from this codebase per govulncheck: `quic-go` v0.59.1, `pgx/v5` v5.9.2, `go-jose/v4` v4.1.4, `oras-go/v2` v2.6.2, `x/text` v0.39.0, `x/net` v0.56.0, plus lockstep `x/*` bumps. All 25 previously reachable findings (19 stdlib, 6 module) now resolve: stdlib ones by go1.26.5, module ones by these upgrades.

### golangci-lint v2 migration
golangci-lint v1 binaries are built with go1.24 and refuse to target Go 1.26, so the config is migrated to the v2 schema (official `golangci-lint migrate`) and the action bumped to v9.3.0 with golangci-lint pinned at v2.12.2. v2 folds `gosimple` into `staticcheck`; staticcheck is restricted to the `S*` checks so lint scope stays identical to what ran before. Enabling the wider SA/ST/QF checks (12 findings today, including a deprecated `oras.Pack` usage) is left as a separate cleanup.

## Acceptance criteria mapping

- *CI rejects unsupported Go lines* — `check-go-toolchain.sh` in CI and (strict) in the release gate
- *Final-artifact provenance records the exact compiler* — the pinned toolchain is embedded in every Go binary's buildinfo (`go version -m <binary>`); full SBOM/provenance is tracked separately in https://github.com/nebari-dev/nebi/issues/211
- *Reachable High/Critical vulnerabilities fail the release* — govulncheck gates fail on **any** reachable finding (stricter than High/Critical); Trivy gates the container image on fixable High/Critical
- *Exceptions* — govulncheck has no ignore file by design (reachable findings should be fixed); Trivy exceptions would go in a `.trivyignore` with a comment carrying owner/expiry — none are needed today

## Verification

- All CI checks green, including the new gates running against this PR itself
- `go build ./cmd/... ./internal/... .` and full `go test -count=1 -tags=e2e ./cmd/... ./internal/...` pass on go1.26.5 (two `cmd/nebi` e2e tests fail identically on unmodified `main` in my local environment, before and after — pre-existing, unrelated)
- `govulncheck ./cmd/... ./internal/... .` and binary-mode scan of an unstripped build → 0 reachable vulnerabilities
- `check-go-toolchain.sh` exercised for all four paths: supported+current (pass), stale patch (warn), stale patch `--strict` (fail), unsupported line (fail)
- `golangci-lint run` (v2.12.2) → 0 issues locally

## Notes for maintainers

- When a new Go patch releases, CI turns yellow (warning) and any release attempt fails until the `toolchain` directive in `go.mod` and the `golang` image digest in `Dockerfile` are bumped — that is the intended "maximum patch latency = zero at release time" policy.
- The wails install pin comment in `desktop.yml` still holds: wails stays at v2.12.0 (matching go.mod); it now builds with Go 1.26, which v2.12.0 supports.


aktech:
I am looking at it now.

.github/workflows/docker.yml
```diff
@@ -143,6 +143,19 @@ jobs:
...
+      # loaded locally.
+      - name: Scan image for vulnerabilities
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
+        with:
+          image-ref: ${{ env.REGISTRY }}/${{ env.PROJECT }}/${{ env.IMAGE_NAME }}:${{ steps.meta.outputs.version }}
+          format: table
+          exit-code: '1'
+          severity: CRITICAL,HIGH
+          ignore-unfixed: true
+
```
aktech: Shouldn't this run between the docker image build and push (we'd need to split that)? so that we don't push the image that fails the vulnerabilities check?

.github/workflows/ci.yml
```diff
@@ -106,15 +106,15 @@ jobs:
...
 
       - name: Create stub frontend/dist for embed
         run: mkdir -p frontend/dist internal/web/dist && touch frontend/dist/.gitkeep internal/web/dist/.gitkeep
 
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@8564da7cb3c6866ed1da648ca8f00a258ef0c802 # v6.5.2
+        uses: golangci/golangci-lint-action@ba0d7d2ec06a0ea1cb5fa41b2e4a3ab91d21278a # v9.3.0
         with:
-          version: latest
+          version: v2.12.2
```
aktech, line 117: We would need to cleanup https://github.com/nebari-dev/nebi/blob/b4d2f1a05692d4c12b31dedd567771dfebbdf0b2/Makefile#L24 to match version of golangci-lint, but not a blocker for this PR.

.github/workflows/docker.yml
tylerpotts: Fixed in e900931. Split the build from the push: the image is now always built locally (`load: true`), the smoke test and trivy scan run against that local image, and a separate `Push image` step pushes the tags only after both pass. Since the pushed bytes are the exact image that was scanned (plain `docker push` of the loaded image, not a rebuild), a failing image can never reach the registry.

.github/workflows/ci.yml
tylerpotts: Went ahead and did it in e900931 since it was a two-line change: `install-tools` and the `lint` fallback install now pin golangci-lint v2.12.2, matching the version used by the CI action.

aktech (approved):

PR #493, opened 2026-08-05
Add secure build environment variables

### LINKED ISSUE #473: [FEAT] -  Improve private package dependency experience
### Problem

I need to create nebi environments using private packages (i.e. private Gitlab repos). The only way to do that currently is by adding my username and PAT to my pixi.toml. I'd like a more secure way of doing this. 

Also, I need to push my environment to a private OCI registry that I set up in nebi. I definitely don't want to push my credentials there where others will pull them down. 

I'm interested in the nebi inside of JHub experience. 

### Proposed solution

I propose that there is some place in the UI that I can specify this information. It seems like being able to add some env vars that I can reference in my pixi.toml would be widely useful - set it in one place, use it in every env build. 

### Alternatives considered

_No response_

### Additional context

_No response_

MUFFANUJ:
## Reference Issues or PRs

Closes #473

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [x] Did you add new tests?

#### Text styling

#### Non-text content

## Any other comments?

Adds per-user build environment variables for private package and environment builds. Values are encrypted at rest and never returned by the API, then injected server-side into Pixi build/install commands while logs and errors are redacted. The implementation also blocks configured secret values from being saved into Pixi artifacts or pushed/published to OCI. This currently uses Nebi’s existing JWT-derived field-encryption key; a separate build-env field encryption secret can be added later if needed.

Remote server `URLs` now require `HTTPS` unless they target localhost. Existing desktop/local configurations that point at non-local `http://` remote servers will stop proxying after the upgrade and should be changed to `https://`.

Also included: a local CLI store fix for symlinked workspace paths, mainly macOS `/tmp` -> `/private/tmp`. On exact path misses, `FindWorkspaceByPath` canonicalizes stored paths and scans local workspace records, so the fallback is O(n) with filesystem calls. This is unrelated to build secrets but fixes existing macOS status/remove e2e flakes.


internal/service/build_env.go
```diff
@@ -0,0 +1,215 @@
...
+// EnsureNoBuildEnvironmentSecretLeak checks artifact contents against the current user's build variables.
+func (s *WorkspaceService) EnsureNoBuildEnvironmentSecretLeak(userID uuid.UUID, contents map[string]string) error {
+	_, secretValues, err := s.buildEnvironmentForUser(userID)
+	if err != nil {
+		return err
+	}
+	return CheckBuildEnvironmentSecretLeak(contents, secretValues)
+}
+
+func normalizeBuildEnvKey(key string) (string, error) {
```
tylerpotts, line 254: **Blocker:** `normalizeBuildEnvKey` accepts any `[A-Za-z_][A-Za-z0-9_]*` name, including `PATH`, `LD_PRELOAD`, `DYLD_INSERT_LIBRARIES`, `HOME`, `PIXI_CACHE_DIR`, `XDG_CACHE_HOME`. `mergeEnvironment` overlays these onto `os.Environ()` for pixi processes that run as the *server* user in team mode, so a user-supplied variable can redirect binary resolution or poison shared caches that other users' builds consume. Users can already run code via package install scripts, so this may not be a brand-new privilege boundary, but cache/loader overrides have cross-user impact. Suggest a deny-list of sensitive names (`PATH`, `HOME`, `LD_*`, `DYLD_*`, `XDG_*`, `PIXI_*`) or an allow-prefix policy, or an explicit written justification for why this is out of threat model.

internal/service/build_env.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// UpsertBuildEnvVar creates or replaces a single current-user build variable.
+func (s *WorkspaceService) UpsertBuildEnvVar(userID uuid.UUID, req BuildEnvVarReq) (*BuildEnvVarResult, error) {
+	key, err := normalizeBuildEnvKey(req.Key)
+	if err != nil {
+		return nil, err
+	}
+	if req.Value == "" {
+		return nil, &ValidationError{Message: "value is required"}
```
tylerpotts: **Question:** no minimum value length. A value like `"6"` or `"name"` will substring-match essentially every `pixi.lock`, so all of that user's builds fail the leak check with a confusing error, and log redaction will shred output with `[REDACTED]`. A minimum length (say 8) would prevent this foot-gun cheaply.

internal/service/build_env.go
```diff
@@ -0,0 +1,215 @@
...
+
+	var envVar models.BuildEnvVar
+	err = s.db.Where("user_id = ? AND key = ?", userID, key).First(&envVar).Error
+	switch {
+	case err == nil:
+		envVar.Value = encryptedValue
+		if err := s.db.Save(&envVar).Error; err != nil {
+			return nil, fmt.Errorf("update build environment variable: %w", err)
+		}
+	case err == gorm.ErrRecordNotFound:
```
tylerpotts: **Question:** two concurrent PUTs of the same new key race between `First` and `Create`; the loser gets a 409 even though PUT-upsert semantics suggest retrying as an update. Minor, but a retry-once-on-unique-violation would make the endpoint idempotent under concurrency.

internal/service/build_env.go
```diff
@@ -0,0 +1,215 @@
...
+
+	result := make([]BuildEnvVarResult, len(vars))
+	for i, v := range vars {
+		result[i] = buildEnvVarToResult(v)
+	}
+	return result, nil
+}
+
+// UpsertBuildEnvVar creates or replaces a single current-user build variable.
+func (s *WorkspaceService) UpsertBuildEnvVar(userID uuid.UUID, req BuildEnvVarReq) (*BuildEnvVarResult, error) {
```
tylerpotts, line 78: **Question:** workspace, group, user, and permission mutations all write audit log entries, but creating/rotating/deleting a credential does not. Registry credential CRUD also skips auditing, so the precedent is ambiguous, but secret rotation is a canonical audit event. Key-only entries (never values) would fit the existing pattern.

internal/service/build_env.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// BuildEnvVarReq holds parameters for creating or updating a build variable.
+type BuildEnvVarReq struct {
+	Key   string
+	Value string
+}
+
+// ListBuildEnvVars returns public metadata for the current user's build variables.
+func (s *WorkspaceService) ListBuildEnvVars(userID uuid.UUID) ([]BuildEnvVarResult, error) {
```
tylerpotts, line 64: **Question (design):** these are user-scoped credentials, not workspace state, yet the CRUD hangs off `*WorkspaceService`. The codebase precedent for exactly this shape (db + encKey credential store) is `RegistryService`. A dedicated `BuildEnvService` passed to the handler and worker would keep `WorkspaceService` from accreting user-profile concerns. Not blocking, but worth considering before the surface grows.

internal/pkgmgr/env.go
```diff
@@ -0,0 +1,76 @@
...
+	"os"
+	"sort"
+	"strings"
+)
+
+type environmentContextKey struct{}
+
+// WithEnvironment attaches additional environment variables to package-manager
+// commands executed with ctx. The input map is copied so callers can mutate it.
+func WithEnvironment(ctx context.Context, env map[string]string) context.Context {
```
tylerpotts, line 14: **Question (design):** passing env through `context.Context` bypasses the `InitOptions`/`InstallOptions`/... structs that are the `PackageManager` interface's established parameter channel. Every exec site must remember `cmd.Env = pkgmgr.CommandEnvironment(ctx)` (8 sites today, all covered), and the contract is invisible in the `PackageManager`/`Executor` interfaces, so a future uv implementation or docker executor will silently drop credentials with no compile-time signal. I can see why ctx was chosen (it crosses both interfaces without signature changes), but at minimum please document the contract on the interfaces; an `Env` field on the options structs would be the pattern-conformant alternative.

internal/worker/worker.go
```diff
@@ -270,6 +391,11 @@ func (w *Worker) executeJob(ctx context.Context, job *models.Job, logWriter io.W
...
 		}
 
+		if err := w.rejectBuildEnvSecretArtifacts(ws, buildEnvSecretValues); err != nil {
+			w.svc.SetWorkspaceStatus(ws.ID, models.WsStatusFailed)
+			return err
+		}
+
 		// List installed packages and save to database
 		if err := w.svc.SyncPackagesFromWorkspace(ctx, ws); err != nil {
 			w.logger.Error("Failed to sync packages", "error", err)
```
tylerpotts: **Question:** job log output and job errors are redacted, but this `w.logger.Error(...)` (and the ones after solve/rollback) logs raw errors to the server log. `SyncPackagesFromWorkspace` runs `pixi list` with the secrets in env, so its errors could embed a secret. Low likelihood, but the redaction boundary is inconsistent; wrapping these through `redactBuildEnvironmentSecrets` would close it.

internal/worker/worker.go
```diff
@@ -227,7 +231,108 @@ func (w *threadSafeWriter) Write(p []byte) (n int, err error) {
...
+	}
+
+	redacted := redactBuildEnvironmentSecrets(err.Error(), secretValues)
+	if redacted == err.Error() {
+		return err
+	}
+	return errors.New(redacted)
+}
+
+func redactBuildEnvironmentSecrets(text string, secretValues []string) string {
```
tylerpotts, line 296: **Stylistic:** the filter/sort-longest-first logic here duplicates what `newBuildEnvironmentLogRedactor` does at construction; the redactor could just call this function (or both could share one helper). Also note a value containing `\n` can span two flushes of the line-buffered redactor and escape redaction; since values are unvalidated free text, consider rejecting control characters in values.

internal/worker/worker.go
```diff
@@ -227,7 +231,108 @@ func (w *threadSafeWriter) Write(p []byte) (n int, err error) {
...
+const buildEnvironmentSecretRedaction = "[REDACTED]"
+
+type buildEnvironmentLogRedactor struct {
+	mu      sync.Mutex
+	writer  io.Writer
+	secrets []string
+	pending []byte
+}
+
+func newBuildEnvironmentLogRedactor(writer io.Writer, secretValues []string) *buildEnvironmentLogRedactor {
```
tylerpotts, line 243: Nice: line-buffering the redactor so a secret split across chunked writes still gets redacted, plus longest-first replacement for overlapping secrets, and the tests cover both cases.

app.go
```diff
@@ -183,8 +184,15 @@ func (a *App) startEmbeddedServer(cfg *config.Config, database *gorm.DB) {
 	}
 	logToFile("startEmbeddedServer: executor initialized")
 
-	// Create service and worker (desktop app uses local mode, no encryption key needed)
-	svc := service.New(database, jobQueue, exec, true, nil, rbac.NewDefaultProvider())
+	encKey, err := nebicrypto.DeriveKey(cfg.Auth.JWTSecret)
+	if err != nil {
+		logToFile(fmt.Sprintf("startEmbeddedServer: encryption key error: %v", err))
```
tylerpotts: **Question:** two things here. (1) On key-derivation failure this silently `return`s, leaving the desktop app running with no embedded server and only a log-file breadcrumb; surfacing the failure would save someone a confusing debugging session. (2) In local mode `cfg.Auth.JWTSecret` is allowed to remain the default `"change-me-in-production"`, so desktop-stored build vars are encrypted with a publicly known key. That is consistent with what `router.go` already does for registry credentials, but AGENTS.md's "local mode: no encryption key needed" is now stale and worth updating so the at-rest guarantee is not oversold.

internal/api/handlers/remote.go
```diff
@@ -31,13 +33,34 @@ func (h *RemoteHandler) getClient() (*cliclient.Client, error) {
...
+		return nil, err
+	}
 	var creds store.Credentials
 	if err := h.db.First(&creds).Error; err != nil || creds.Token == "" {
 		return nil, fmt.Errorf("not authenticated with remote server")
 	}
 	return cliclient.New(cfg.ServerURL, creds.Token), nil
 }
 
+func validateRemoteServerURL(rawURL string) error {
```
tylerpotts, line 46: **Question (scope/behavior change):** enforcing https on connect *and* on every stored-config proxy call is defensible hardening now that the proxy forwards secret values, but it changes behavior for all existing remote features: a previously working `http://` server URL stored in config starts failing with 503s after upgrade. This deserves a callout in the PR description (or its own PR) so operators are not surprised.

internal/store/workspace.go
```diff
@@ -37,6 +69,14 @@ func (s *Store) FindWorkspaceByPath(path string) (*LocalWorkspace, error) {
 	return &ws, nil
 }
 
+func canonicalizeExistingPath(path string) (string, bool) {
```
tylerpotts, line 72: **Question (scope):** the symlink canonicalization looks like an unrelated bugfix (macOS `/tmp` symlink e2e flake; I confirmed it fixes `TestE2E_StatusJSON`/`TestE2E_WorkspaceRemove` locally) bundled into a secrets PR. It also adds a fallback that scans every stored workspace and calls `EvalSymlinks` on each, making `FindWorkspaceByPath` O(n) with syscalls on the miss path. Fine for a local CLI store, but worth splitting out or at least mentioning in the PR body.

frontend/src/hooks/useBuildEnv.ts
```diff
@@ -0,0 +1,55 @@
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
+import { type BuildEnvTarget, buildEnvApi } from '@/api/buildEnv';
+import type { BuildEnvVar, UpsertBuildEnvVarRequest } from '@/types';
+
+const buildEnvVarsQueryKey = (target: BuildEnvTarget) =>
```
tylerpotts, line 5: **Stylistic:** the query keys are asymmetric (`['remote', 'build-env-vars']` vs `['build-env-vars', 'local']`). If the `remote` prefix is intentional so disconnect invalidation catches it, a one-line comment would help; otherwise make them symmetric.

internal/worker/worker_test.go
```diff
@@ -249,6 +272,170 @@ func TestExecuteJob_UpdateSetsWorkspaceFailedOnSolveError(t *testing.T) {
...
+	var stored models.Workspace
+	if err := db.First(&stored, "id = ?", ws.ID).Error; err != nil {
+		t.Fatalf("reload workspace: %v", err)
+	}
+	if stored.Status != models.WsStatusFailed {
+		t.Fatalf("expected workspace status failed, got %q", stored.Status)
+	}
+}
+
+func TestExecuteJob_RedactsBuildEnvironmentSecretsFromLogsAndErrors(t *testing.T) {
```
tylerpotts, line 378: Nice test: asserting the secret is absent from logs *and* the returned error, and that the redaction marker is present, proves the property rather than the implementation.

tylerpotts (changes requested):
Thanks for this. The implementation is strong overall: the layering matches existing resources, encryption reuses the exact `RegistryService` convention, values are never serialized anywhere, and the tests prove the security properties (no secret in API responses, logs, error strings, or recorded job errors) rather than just the happy path. I verified locally that all touched Go packages, the e2e suite, and the frontend hook tests pass, and that all 8 pixi exec sites get the env injection.

This resolves both halves of https://github.com/nebari-dev/nebi/issues/473: credentials set once and injected into every build, and blocked from being persisted or published to OCI.

**Requesting changes for one issue:** there is no restriction on variable *names*, so a user can override process-level environment (`PATH`, `LD_PRELOAD`, `PIXI_CACHE_DIR`, `HOME`, ...) for pixi processes that run as the server user in team mode. Details inline at `internal/service/build_env.go`. Everything else below is a question or a nit.

One verification caveat: the tests use stub pixi binaries, so what is proven is that the variables reach the pixi process environment. A manual smoke test against a real private index/repo (pixi/uv/git consuming the vars) would be good before closing the issue.

tylerpotts:
A few general notes to go with the review (details are inline there):

**Verification done locally**
- `go build` / `go vet` clean; all touched Go packages pass (`service`, `worker`, `executor`, `pkgmgr`, `pkgmgr/pixi`, `api/handlers`, `store`, `crypto`), plus the `-tags e2e` suite and the frontend `useBuildEnv` tests.
- Confirmed all 8 `exec.CommandContext` sites in `internal/pkgmgr/pixi/pixi.go` and `internal/executor/local.go` receive the injected environment.
- Confirmed the publish path only ever pushes `pixi.toml` + `pixi.lock`, so the leak check covers the full artifact surface.
- The symlink change fixes real environment-specific e2e failures on macOS (`TestE2E_StatusJSON`, `TestE2E_WorkspaceRemove`), verified before/after.

**Architecture fit**
The change follows the existing conventions well: handler -> service -> model layering, the `RegistryService` field-encryption pattern (`enc:v1:`, encKey on the service struct, values never serialized), the service error taxonomy, regenerated (not hand-edited) swagger, cliclient mirroring for the desktop proxy, and the api-client/hook/page structure on the frontend.

**Docs suggestion**
Two now-load-bearing realities are undocumented and worth adding to AGENTS.md, ideally in this PR:
1. "Local mode vs. team mode" still says local mode needs no encryption key; both entry points now derive a field-encryption key from `auth.jwt_secret`, and in local mode that secret may be the default, so at-rest encryption of credentials is only meaningful in team mode with a real secret.
2. The new pkgmgr contract: any code that shells out to a package manager must set `cmd.Env = pkgmgr.CommandEnvironment(ctx)` or user build variables silently will not apply. The first uv/docker executor implementation will miss this otherwise.

**Before closing https://github.com/nebari-dev/nebi/issues/473**
The tests use stub pixi binaries, so they prove the variables reach the pixi process. One manual smoke test against a real private index or git repo (pixi/uv/git actually consuming the credentials) would close the loop on the original use case.

internal/service/build_env.go
MUFFANUJ: Planning to address this by replacing unrestricted env var names with an admin-configured allow-prefix policy. The default allowed prefix is `NEBI_`, so process-control variables such as `PATH`, `HOME`, `LD_*`, `DYLD_*`, `XDG_*`, and `PIXI_*`, etc are rejected by default. Admins can explicitly add prefixes for their deployment when needed.

internal/pkgmgr/env.go
MUFFANUJ: Agreed this is user-scoped rather than workspace state. I kept it on `WorkspaceService` for this PR to minimize the change surface because the worker already depends on `WorkspaceService` for `build`/`persist`/`publish` checks. If this surface grows, I agree it should move to a dedicated `BuildEnvService` similar to `RegistryService`.

internal/api/handlers/remote.go
MUFFANUJ: Update the description

internal/store/workspace.go
MUFFANUJ: Updated!

internal/service/build_env.go
```diff
@@ -0,0 +1,369 @@
...
+	return values
+}
+
+// ReadWorkspaceArtifactContents reads the Pixi files that can be persisted or published.
+func ReadWorkspaceArtifactContents(wsPath string) (map[string]string, error) {
+	contents := make(map[string]string, len(workspaceArtifactFilenames))
+	for _, filename := range workspaceArtifactFilenames {
+		content, err := os.ReadFile(filepath.Join(wsPath, filename))
+		if err != nil {
+			return nil, fmt.Errorf("read %s: %w", filename, err)
```
tylerpotts, line 212: **Blocker:** this hard-fails when an artifact file does not exist, and the worker calls `rejectBuildEnvSecretArtifacts` after every create job whenever the user has at least one build variable (`internal/worker/worker.go:389`). The bare-create path (the `default:` branch in `internal/executor/local.go`) runs only `pixi init`, which writes `pixi.toml` but never a `pixi.lock`. Concrete failure: save `GITHUB_TOKEN`, create a workspace without supplying a `pixi.toml`, and the create job errors with `read pixi.lock: ... no such file or directory`, marking the workspace failed. The worker tests miss it because `fakeExecutor.CreateWorkspace` unconditionally writes both files. A missing artifact cannot leak a secret, so skipping files that do not exist (`os.IsNotExist`) is safe here. This was in round 1's code and I missed it then; flagging now because it ships as a latent regression triggered by adopting the feature.

internal/service/build_env.go
```diff
@@ -0,0 +1,369 @@
...
+	for _, prefix := range s.buildEnvAllowedPrefixes {
+		allowed = append(allowed, prefix+"*")
+	}
+	sort.Strings(allowed)
+	return &ValidationError{Message: fmt.Sprintf("key %q is not allowed; allowed names or prefixes: %s", key, strings.Join(allowed, ", "))}
+}
+
+func normalizeBuildEnvPolicy(policy BuildEnvPolicy) (map[string]struct{}, []string) {
+	if len(policy.AllowedNames) == 0 && len(policy.AllowedPrefixes) == 0 {
+		policy = defaultBuildEnvPolicy
```
tylerpotts, line 296: **Blocker:** an explicitly empty policy silently falls back to the defaults (here and again at the bottom of this function). An admin who sets `allowed_names: []` and `allowed_prefixes: []` in `config.yaml` to disable user build variables instead gets `GITLAB_TOKEN`/`GITHUB_TOKEN`/`GH_TOKEN` plus `NEBI_*` allowed, and there is no other off switch (the routes are always registered). `docs/docs/server-setup.md` says the server only accepts names allowed by admin configuration and never mentions the fallback. The codebase precedent for operator misconfiguration is fail-loud at `config.Load` (`validateTeamModeJWTSecret`, `normalizeRegistries`), not silent substitution. Suggest: apply the default only when the `build_env` section is absent (`v.IsSet` in `config.Load`) and treat explicitly-empty as deny-all. That also lets you delete `defaultBuildEnvPolicy` here, which duplicates the viper defaults in `internal/config/config.go` and has to be kept in sync by hand.

internal/service/workspace.go
```diff
@@ -19,17 +19,33 @@ import (
...
+	isLocal                 bool
+	encKey                  []byte
+	buildEnvAllowedNames    map[string]struct{}
+	buildEnvAllowedPrefixes []string
 }
 
 // New creates a new WorkspaceService.
-func New(db *gorm.DB, q queue.Queue, exec executor.Executor, isLocal bool, encKey []byte, rbacProvider rbac.Provider) *WorkspaceService {
-	return &WorkspaceService{db: db, queue: q, executor: exec, isLocal: isLocal, encKey: encKey, rbac: rbacProvider}
+func New(db *gorm.DB, q queue.Queue, exec executor.Executor, isLocal bool, encKey []byte, rbacProvider rbac.Provider, buildEnvPolicies ...BuildEnvPolicy) *WorkspaceService {
```
tylerpotts, line 33: **Stylistic:** this is the only variadic constructor in the repo, and all three production call sites (`app.go`, `internal/server/server.go`, `internal/api/router.go`) pass the policy explicitly; only tests omit it. A required seventh parameter would make a forgotten policy a compile error instead of a silent default, would remove the need for the empty-policy fallback in `normalizeBuildEnvPolicy`, and matches the codebase idiom of explicit positional deps (`worker.New` takes a `nil` valkey client). Elements beyond the first are also silently ignored.

internal/api/handlers/build_env.go
```diff
@@ -0,0 +1,95 @@
...
+// @Description Stores a current-user build environment variable. Values are encrypted at rest and are never returned.
+// @Tags build-env-vars
+// @Accept json
+// @Produce json
+// @Security BearerAuth
+// @Param request body BuildEnvVarRequest true "Build environment variable"
+// @Success 200 {object} service.BuildEnvVarResult
+// @Failure 400 {object} ErrorResponse
+// @Failure 401 {object} ErrorResponse
+// @Failure 409 {object} ErrorResponse
```
tylerpotts, line 50: **Stylistic:** with the upsert now retrying on unique violation, the service never returns a `ConflictError` on this path, so this documented 409 is unreachable. Drop the annotation and regenerate swagger (CI now fails on swagger drift).

internal/service/build_env.go
```diff
@@ -0,0 +1,369 @@
...
+	return names, prefixes
+}
+
+func validateBuildEnvValue(value string) error {
+	trimmed := strings.TrimSpace(value)
+	if trimmed == "" {
+		return &ValidationError{Message: "value is required"}
+	}
+	if utf8.RuneCountInString(trimmed) < minBuildEnvValueLength {
+		return &ValidationError{Message: fmt.Sprintf("value must be at least %d characters", minBuildEnvValueLength)}
```
tylerpotts, line 334: **Question:** the minimum of 8 contradicts the PR's own documented example. `config.yaml.example` and `docs/docs/server-setup.md` both recommend allowing `UV_INDEX_USERNAME`, but a username like `deploy` cannot be stored. The minimum length (my round 1 ask) makes sense for tokens; applied uniformly it blocks the documented use case. Consider lowering it, or keeping 8 for redaction safety and dropping the username examples from the docs.

internal/service/build_env.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+	sort.Strings(prefixes)
+	if len(names) == 0 && len(prefixes) == 0 {
+		return normalizeBuildEnvPolicy(defaultBuildEnvPolicy)
+	}
+	return names, prefixes
+}
+
+func validateBuildEnvValue(value string) error {
+	trimmed := strings.TrimSpace(value)
```
tylerpotts, line 329: **Stylistic:** the value is length-validated on the trimmed string but stored untrimmed, so `"  token123 "` is stored with the padding and injected verbatim, while a pasted trailing newline is rejected as a control character rather than trimmed. Either store the trimmed value or validate the raw one, consistently.

internal/service/build_env.go
```diff
@@ -0,0 +1,369 @@
...
+	var vars []models.BuildEnvVar
+	if err := s.db.Where("user_id = ?", userID).Order("key ASC").Find(&vars).Error; err != nil {
+		return nil, nil, fmt.Errorf("fetch build environment variables: %w", err)
+	}
+
+	env := make(map[string]string, len(vars))
+	for _, v := range vars {
+		if err := s.validateBuildEnvKeyAllowed(v.Key); err != nil {
+			return nil, nil, &ValidationError{
+				Message: fmt.Sprintf("build environment variable %q is not allowed by the configured build environment policy; delete it before running builds", v.Key),
```
tylerpotts, line 172: **Question:** when a stored var no longer passes a tightened policy, this fails `buildEnvironmentForUser`, which also fails `EnsureNoBuildEnvironmentSecretLeak`, so workspace create, `SavePixiToml`, and push all return 400 until the user deletes the var. Fail-closed is defensible, the message is actionable, and delete still works for reserved keys (nice test for that), but the error surfaces on operations that look unrelated to build secrets. Confirming the blast radius is intended.

internal/service/build_env.go
```diff
@@ -0,0 +1,369 @@
...
+		}
+	case errors.Is(err, gorm.ErrRecordNotFound):
+		auditAction = audit.ActionCreateBuildEnvVar
+		envVar = models.BuildEnvVar{
+			UserID: userID,
+			Key:    key,
+			Value:  encryptedValue,
+		}
+		if err := s.db.Create(&envVar).Error; err != nil {
+			if isUniqueConstraintError(err) {
```
tylerpotts, line 112: **Question:** no test exercises this retry path. The string matching (`unique`/`duplicate`) covers sqlite and Postgres today, but a test pinning the create-conflict-then-update behavior would keep it honest.

internal/worker/worker.go
```diff
@@ -438,14 +588,14 @@ func (w *Worker) executeJob(ctx context.Context, job *models.Job, logWriter io.W
 // here must not fail the whole job: that would hide a change that
 // actually succeeded. Instead it's logged and recorded as a failed env
 // install, which surfaces as install_status = install_failed.
-func (w *Worker) maybeReinstallEnv(ctx context.Context, ws *models.Workspace, wasInstalled bool, logWriter io.Writer) error {
+func (w *Worker) maybeReinstallEnv(ctx context.Context, ws *models.Workspace, wasInstalled bool, logWriter io.Writer, secretValues []string) error {
```
tylerpotts, line 591: **Question:** `pixi install` runs without `--frozen`/`--locked` (`internal/executor/local.go`), so it can re-solve and rewrite `pixi.lock` when the lock is out of sync with the manifest, yet `JobTypeEnvInstall` and this reinstall path are the only lock-mutating paths that never re-run `rejectBuildEnvSecretArtifacts` afterward. Is that intentional (lock considered already-checked), or should the check run after any install too?

app.go
```diff
@@ -92,6 +95,20 @@ func (h *appHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	h.app.router.ServeHTTP(w, r)
 }
 
+func (a *App) failEmbeddedServerStartup(logMessage string) {
+	logToFile(logMessage)
+	a.router = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		http.Error(w, "Nebi API failed to start; see /tmp/nebi-startup.log", http.StatusServiceUnavailable)
```
tylerpotts, line 101: **Stylistic:** this user-facing message hardcodes `/tmp/nebi-startup.log`, and `logToFile` silently no-ops on Windows where that path is never written, so Windows desktop users get pointed at a file that does not exist. Pre-existing pattern, but the new fallback handler bakes it into UI-visible text.

tylerpotts (changes requested):
Thanks for the thorough response round. Everything from my first review is addressed: the allow-list policy with a reserved deny-list that runs before the allow-list (so an admin cannot accidentally allow `PATH`), min value length plus control-character rejection (which also closes the newline-across-flushes redactor gap), the upsert retry on unique violation, audit logging with key-only metadata, server-log redaction at all logger sites, the startup fallback handler with tests (which fixes a real latent hang where a startup error left `a.ready` unclosed forever), and the docs/AGENTS.md updates. The redactor consolidation is better than round 1: deterministic longest-first ordering with an overlapping-secrets test.

Verified locally at 2d3765b: `go build` / `go vet` clean, and tests pass for `service`, `worker`, `config`, `pkgmgr`, `executor`, `api/handlers`, `store`, and the root package (`app_test.go`).

**Requesting changes for two issues found this round** (details inline at `internal/service/build_env.go`):

1. Any user with a stored build variable can no longer create a bare workspace: the post-create leak check hard-fails on the missing `pixi.lock` that `pixi init` never writes. This one is on me, the call was in round 1's code and I missed it then.
2. An explicitly empty policy (`allowed_names: []` and `allowed_prefixes: []`) silently falls back to the default allow-list, so there is no way for an admin to disable user build variables, and the silent substitution contradicts the fail-loud precedent in `config.Load` (`validateTeamModeJWTSecret`, `normalizeRegistries`).

Also note: **the branch currently has merge conflicts with `main`**, so it needs a merge/rebase before it can land regardless.

Smaller items inline: the swagger `@Failure 409` annotation is now unreachable after the upsert retry (CI fails on swagger drift, so it needs a regen), the min length of 8 contradicts the documented `UV_INDEX_USERNAME` example, the value is validated trimmed but stored untrimmed, and the variadic `BuildEnvPolicy` parameter on `service.New` is the only variadic constructor in the repo and exists only for test convenience.

tylerpotts:
@MUFFANUJ There's also a few conflicts to resolve now

MUFFANUJ:
Thanks for checking @tylerpotts! I see that the issue is marked with `needs:discussion` hence I'm not sure if we are planning to move forward with this PR or not yet at this moment.

tylerpotts:
> Thanks for checking @tylerpotts! I see that the issue is marked with `needs:discussion` hence I'm not sure if we are planning to move forward with this PR or not yet at this moment.

Per @pmeier he will take over reviewing this

MUFFANUJ:
We discussed this in the meeting, and we plan to have a more organized plan since this might not be what the project is aiming for!

Issue #445, opened 2026-07-14
[security] Untrusted environment builds run in a shared worker with DB credentials and a shared multi-tenant volume


dcmcand:
## Summary

Nebi intentionally lets workspace writers submit manifests and install packages, which by design executes package/build code. The problem is **where** that code runs: the shared worker holds the database DSN (with password), mounts a shared multi-tenant environment volume, and uses the same ServiceAccount as the API, with no isolation or hardening. Untrusted build code therefore runs next to sensitive credentials and other tenants' data. (CWE-250 Execution with Unnecessary Privileges; related CWE-653 Improper Isolation.)

## Where it is (verified against `main` @ `15f66d4`)

- The executor writes a caller-supplied `pixi.toml` verbatim and runs `pixi install -v`: `internal/executor/local.go:136,166`.
- The package API appends caller-provided package strings directly into the `pixi add -v` argv with **no `--` separator**: `internal/pkgmgr/pixi/pixi.go:200-203`. This is argument injection (a value like `--config`/`-e` is parsed as a pixi flag); combined with `pixi.toml` source/build backends it is arbitrary build-code execution. Handler passes `Packages` through with only `binding:"required"` (`internal/api/handlers/workspace.go:600`).
- The worker runs up to **10 concurrent jobs** per process (hardcoded): `internal/worker/worker.go:40,50`.
- The worker Deployment: receives the Postgres DSN incl. password inline in env (`chart/templates/deployment-worker.yaml:38`); connects to Valkey (`:52-57`); mounts the shared `ReadWriteMany` environment PVC also mounted by the API (`:70-75`, `chart/values.yaml:108`); uses the same ServiceAccount as the API (`:21`); does not set `automountServiceAccountToken: false` (absent from all templates); has no rendered pod `securityContext`; and there is no default-deny egress NetworkPolicy anywhere in `chart/templates/`.

> Note: the chart currently fails to render at all (`chart/templates/_helpers.tpl` defines `darb.*` helpers while templates call `nebi.*`). The isolation gaps above are accurate at the chart-source level and take effect once that render bug is fixed.

## Impact

- Build code can read the worker's environment, including the database DSN, and reach the shared environment volume, the queue, the network, and any mounted ServiceAccount token.
- Extraction of credentials present in the worker environment; modification of another workspace's files via the shared volume; queue/job tampering; access to cloud metadata or internal services through unrestricted egress.

## Remediation

[...]

PR #495, opened 2026-08-05
Sandbox untrusted environment builds

tylerpotts:
## Reference Issues or PRs

Part of https://github.com/nebari-dev/nebi/issues/445

Companion work: https://github.com/nebari-dev/nebi/pull/465 fixed the pixi argument injection, and https://github.com/nebari-dev/nebi-pack/pull/45 hardens the Helm chart. This PR addresses the core finding those two did not: untrusted build code executing next to credentials and other tenants' files.

- [ ] Documentation Update
- [ ] Build related changes

## Summary

Environment builds execute untrusted code: a `pixi.toml` can declare source or build backends that run arbitrary commands during a solve or install. Today those commands inherit the full server environment and can reach every tenant's files. This PR closes both halves at the process level.

**Environment scrubbing**, in every mode on every platform. Build subprocesses receive only an allowlist (`PATH`, `HOME`, `TMPDIR`, locale, TLS trust, proxy variables). `NEBI_DATABASE_DSN`, `NEBI_AUTH_JWT_SECRET`, queue addresses, and registry credentials never reach build code. Previously no `exec.Cmd` in `internal/executor` or `internal/pkgmgr` set `cmd.Env` at all.

**Landlock confinement**, on Linux. A hidden `nebi sandbox-exec` subcommand applies a Landlock ruleset to itself and then `syscall.Exec`s the real build command, so pixi and everything it spawns inherit the confinement. A build can read and write only its own workspace directory, read a specific set of system paths, and open TCP connections only to configured ports. The database and queue become unreachable from build code even though they share the pod network.

Confinement needs no extra privileges. It works with `runAsNonRoot`, all capabilities dropped, and no user namespaces, which is what makes it compatible with the chart hardening in nebari-dev/nebi-pack#45.

## Configuration

```yaml
sandbox:
  mode: strict            # strict | permissive | off
  allowed_ports: [80, 443]
  build_timeout: 30m
```

`NEBI_SANDBOX_MODE`, `NEBI_SANDBOX_ALLOWED_PORTS`, `NEBI_SANDBOX_BUILD_TIMEOUT`. Mode defaults to `strict` in team mode and `off` in local/desktop mode.

| mode | when the kernel cannot confine |
|---|---|
| `strict` | the job fails with a message naming `NEBI_SANDBOX_MODE` |
| `permissive` | a warning goes to the build log, the build runs unconfined |
| `off` | nothing is attempted; environment scrubbing still applies |

## Kernel requirements

| Landlock ABI | kernel | what you get |
|---|---|---|
| v1 | 5.13+ | filesystem confinement, but cross-directory renames fail with `EXDEV` |
| v2 | 5.19+ | adds `refer`, which package managers need to stage downloads. Practical minimum |
| v4 | 6.7+ | adds TCP restriction. Feature complete |

Below v4 the network is unrestricted with a warning rather than a build failure, even in strict mode.

## Two operator traps, both documented

- `allowed_ports: []` denies **all** TCP, it does not mean unrestricted. Note that the environment variable cannot express this: viper treats an empty env var as unset, so `NEBI_SANDBOX_ALLOWED_PORTS=""` silently yields the default. Use the YAML form.
- Upgrading an existing team deployment starts enforcing confinement and the build timeout where neither existed. Builds on kernels older than 5.13 will fail until the operator sets `permissive` or `off`.

## Notable things found while building this

Three of these were caught only because the implementation was verified against a real kernel rather than reasoned about:

- **Landlock denies reparenting in every domain unless `refer` is granted, and `refer` can only be granted in the domain that handles it.** A filesystem ruleset followed by a stacked network ruleset therefore re-denies renames regardless of what the filesystem domain allowed. Package managers rename staged downloads into their cache constantly, so this would have broken every Linux build. `Confine` is now a single combined ruleset with a v4 → v2 → v1 ladder. Note for anyone testing this: GNU `mv` falls back to copy-and-unlink on `EXDEV` and reports success, so only a raw `rename(2)` exposes it.
- **Three service-layer `pixi list` calls bypassed the executor entirely** (`syncPackagesFromDisk`, `SyncPackagesFromWorkspace`, `CreateVersionSnapshot`), running against user-controlled workspaces with the full server environment. Closed by adding `ListPackages` to the `Executor` interface, which also fixes a pre-existing bug where those paths ignored `config.PackageManager.PixiPath`.
- **Granting read on `/etc` would have re-exposed the secrets the allowlist removes**, since nebi searches `/etc/nebi/` for `config.yaml` and both `database.dsn` and `auth.jwt_secret` are file-loadable. Only the TLS trust stores and specific resolver files are granted.
- **`config.package_manager.pixi_path` was being ignored** by package listing and version snapshots, which resolved pixi from PATH instead and would silently auto-install one if absent. Routing those through the executor fixed it. This surfaced as an e2e test failure: that test only passed on main because the real pixi it accidentally invoked writes `pixi.lock` as a side effect of `pixi list`, which the version snapshot then read. The test was asserting on a bug.
- Out-of-range ports were silently truncated by `uint16(port)`, so `70000` would have opened port 4464 while the intended port stayed closed. Validated at both the config and ruleset layers.

## Testing

- [x] Did you add new tests?

Unit tests cover the environment allowlist, argv construction, mode resolution and validation, HOME scoping on both sides of the active/off branch, setup-failure detection, and the Landlock rule assembly.

`internal/sandbox/confine_test.go` is the acceptance test issue 445 asks for. It runs a deliberately malicious build backend (`testdata/probe`) confined and asserts it cannot read the planted DSN from the environment, cannot read or write a sibling workspace, cannot read `/etc` outside the allowlist, and cannot connect to a stand-in database port, while a legitimate build in its own workspace succeeds and cross-subdirectory renames still work. Negative cases assert on the specific errno, so a missing file or broken binary cannot masquerade as containment.

The test is `//go:build linux` and skips when the kernel reports no Landlock support. **It is confirmed enforcing in CI**, not skipping: the run logs `kernel Landlock ABI version 7` and all eight subtests pass, including the positive controls (a legitimate write and a cross-subdirectory rename succeed, and an allowed port connects) that stop the negative cases from passing vacuously.

## Two bugs this work surfaced in existing code

Neither is fixed here, and both predate this branch.

**A real race in the create-job API contract.** `internal/worker/worker.go` sets the workspace status to `ready` *before* calling `CreateVersionSnapshot`. A client that polls until `ready` and immediately publishes can legitimately get `400 Workspace has no versions to publish`. The 250ms poll interval in the e2e test hides it. Moving the snapshot above the status flip looks like a three-line fix.

**Snapshot failures are swallowed.** When `CreateVersionSnapshot` fails it only logs. The workspace still goes `ready` with zero versions, and the user learns nothing until a later publish fails with a message that does not name the real cause.

Separately, `cmd/nebi`'s e2e suite still shells out to the real pixi in most tests, which makes it slow and dependent on PATH and the network.

## Not in scope

These remain open under issue 445 and should keep it open after this merges:

- A Kubernetes Job-per-build executor. That is infrastructure-level isolation and needs client-go, chart work, queue ack/retry, and log shipping.
- Queue visibility-timeout and ack semantics. A worker dying mid-build still leaves a job `running` forever, which is pre-existing.
- Per-workspace volumes in nebi-pack.
- Killing the whole process group on timeout. `exec.CommandContext` signals only the direct child, so pixi's grandchildren can outlive a timeout. They stay Landlock-confined, so this is resource cleanup rather than containment, and the fix needs a platform split for the Windows desktop build.
- Private conda channel credentials. Because `HOME` is redirected when the sandbox is active, `$HOME/.rattler/credentials.json` does not reach confined builds. Teams using authenticated channels need `off` today. The right fix is a per-job credential file scoped to the workspace's declared channels, never environment inheritance.

## Documentation

`config.yaml.example` documents every key, and `docs/docs/server-setup.md` gains a "Build Sandbox" section covering the two layers, the mode and ABI tables, both operator traps, the per-workspace cache trade-off, and the known limitations.

#### Text styling

#### Non-text content

- [x] All content is represented as text (for example, images need alt text, and videos need captions or descriptive transcripts).
- [x] If there are emojis, there are not more than three in a row.
- [x] Don't use [flashing GIFs or videos](https://www.w3.org/TR/UNDERSTANDING-WCAG20/seizure-does-not-violate.html).
- [x] If the content were to be read as plain text, it still makes sense, and no information is missing.


MUFFANUJ:
This picked up some conflicts!

MUFFANUJ:
This has picked up conflicts; let's try to merge this soon. Maybe #513 could land better after this?

PR #496, opened 2026-08-07
Update installation docs for windows

pavithraes:
* Add windows installation instruction
* Condense advanced options
* Remove explicit `nebi-desktop` from conda-forge (since it is bundled with the `nebi` package)
* Minor stylistic clean up

### DIFF
--- a/docs/docs/installation.md
+++ b/docs/docs/installation.md
@@ -20,9 +20,7 @@ pixi global install nebi
 
 ## Installation script
 
-:::note
-Supported platforms: **macOS** and **Linux**
-:::
+### Linux & MacOS
 
 This installs the latest release of `nebi` to `~/.local/bin` (CLI and desktop app):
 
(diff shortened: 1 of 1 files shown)


docs/docs/installation.md
```diff
@@ -36,44 +34,32 @@ Make sure it's on your `PATH`:
...
-### Install only CLI
+### Windows (Powershell)
 
 ```bash
-curl -fsSL https://nebi.nebari.dev/install.sh | bash
+irm https://nebi.nebari.dev/install.ps1 | iex
 ```
 
-### Install a specific version
+### Advanced options
```
jbouder: I believe these advanced options only work with curl and not the the tool used here on windows. Suggest moving advanced options under the Linux/Mac section

jbouder (changes requested):

docs/docs/installation.md
pavithraes: Indeed, `install.ps1` defines slightly different option names. Updated now. Thanks!

jbouder (approved):

PR #501, opened 2026-08-12
Fix: White screen flash every ~5s when configured remote server is unreachable

### LINKED ISSUE #217: Bug: White screen flash every ~5s when configured remote server is unreachable
## Bug Description

In the desktop app (local mode), if a remote server is configured but becomes unreachable (server down, network changed, etc.), the UI produces a white screen flash approximately every 5 seconds. The flashing stops once the user navigates to Settings and clicks **Disconnect**.

## Steps to Reproduce

1. Start nebi in local mode
2. Connect to a remote server via Settings
3. Switch to Remote view so remote workspaces are visible
4. Stop the remote server (or disconnect from its network)
5. Observe: the screen flashes white every ~5 seconds

**Workaround:** Go to Settings → Disconnect from the remote server.

## Root Cause Analysis

Several React Query hooks poll the remote server on a fixed interval that continues firing even when the server is unreachable:

- `useRemoteWorkspaces()` — `refetchInterval: 5000` (`frontend/src/hooks/useRemote.ts:40`)
- `useRemoteJobs()` — `refetchInterval: 5000` (`frontend/src/hooks/useRemote.ts:103`)
- `useRemoteServer()` — `refetchInterval: 10000` (`frontend/src/hooks/useRemote.ts:9`)

When the server is unreachable:

1. The 5-second polls fail, returning `undefined` data
2. The query client retries once (`retry: 1` in `frontend/src/lib/queryClient.ts:7`), then errors
3. Each error/retry cycle causes a re-render where remote data is momentarily `undefined`
4. In `Workspaces.tsx:81-84`, when `remoteWorkspaces` is undefined/empty, the displayed list flips to `[]`, causing a visible flash
5. This repeats every 5 seconds indefinitely

**Why Disconnect fixes it:** `useDisconnectServer` invalidates all `['remote']` queries (`useRemote.ts:30`), which clears the polling hooks and stops the cycle.

## Suggested Fix

Options (not mutually exclusive):

1. **Stop polling on error**: Use `refetchInterval: (query) => query.state.status === 'error' ? false : 5000` to pause polling when the server is unreachable, resuming only after a successful manual reconnect.
2. **Pause polling when connection status is not `connected`**: The `useRemoteServer` hook already tracks connection status. Hooks like `useRemoteWorkspaces(enabled)` receive `enabled = isRemoteConnected`, but `refetchInterval` fires regardless of the `enabled` flag once the query has been activated. The interval should be conditional: `refetchInterval: enabled ? 5000 : false`.
3. **Graceful error state in UI**: Instead of rendering an empty list on query error, show a banner like "Remote server unreachable — retrying..." and preserve the last successful data (`placeholderData: keepPreviousData`).

## Affected Components

- `frontend/src/hooks/useRemote.ts` — polling intervals
- `frontend/src/pages/Workspaces.tsx` — workspace list rendering
- `frontend/src/components/jobs/Jobs.tsx` — jobs list rendering
- `frontend/src/lib/queryClient.ts` — retry configuration

jbouder:
## Reference Issues or PRs

Fixes #217 

- [ ] Documentation Update
- [ ] Build related changes

### Summary

When a configured remote server is unreachable, the remote-view pages no longer flash the full-page loading spinner every poll cycle. Instead they show an "unreachable" banner (Workspaces, Jobs, and the Admin Dashboard), keep polling on a slower 30s backoff so the UI self-heals once the server comes back, and only show the full-page spinner on the very first load.

> **Scope note:** this PR also changes the app-wide TanStack Query `networkMode` default. The query client now starts with `networkMode: 'always'` and `modeStore` switches it per app mode after `/version` resolves: `'always'` in local (desktop) mode — the loopback backend stays reachable even when the OS reports offline — and `'online'` in team mode, which pauses queries/mutations while offline. Every existing and future query/mutation inherits this default, so it deserves changelog visibility beyond the banner fix. The invariants are documented in AGENTS.md.
>
> Bootstrap behavior also changes: `fetchMode` now retries `/version` (3 attempts, 300ms apart) to cover the desktop embedded-server startup race, which can delay the team-mode fallback by up to ~600ms when `/version` never answers, and the failure path only downgrades `networkMode` to `'online'` when the Wails runtime is absent.
>
> Follow-up: migrating the remaining remote-view pages (Registries + admin pages) to the shared `useRemoteView()` gating is tracked in #504.

## Testing

- [x] Did you add new tests?

#### Text styling

#### Non-text content

## Any other comments?
Screenshot 2026-08-12 at 1 11 58 PM


tylerpotts (changes requested):
Reviewed in a local worktree: full frontend suite passes (194/194), typecheck and lint clean. The main thing I checked first is that the banner cannot appear when no remote server is configured, and that holds structurally: `remoteUnreachable` requires `isRemoteConnected`, which requires `serverStatus?.status === 'connected'`, and the backend (`internal/api/handlers/remote.go`) only reports `connected` when both a server URL and token are stored. On top of that, `useRemoteJobs`/`useRemoteWorkspaces` are `enabled: isRemoteConnected`, so with no server configured the error-producing queries never run at all. Nice double gate.

That said, I'm requesting changes because a few findings undermine the fix itself:

### Blocking

**1. Polling never recovers after a single error** (`frontend/src/hooks/useRemote.ts` `pollUnlessErrored`)
Once a poll fails (retry: 1), `refetchInterval` returns `false` permanently. `/api/v1/remote/server` is a local DB read that always reports `connected`, `refetchOnWindowFocus` is globally false, and only the connect/disconnect/create/delete mutations invalidate `['remote']`. So if the remote server restarts briefly, the banner tells the user to check that the server is running, they restart it, and the banner never clears until they navigate away or reconnect in Settings. An error backoff (e.g. return `30000` when errored instead of `false`) fixes the flash without losing self-healing.

**2. The `!remoteError` spinner guard is dead during any real refetch** (`Jobs.tsx`, `Workspaces.tsx`)
Verified against @tanstack/query-core 5.90.5: when an errored query with no data refetches, `fetchState` resets `error: null, status: 'pending'`, so `isError` flips false and the full-page spinner flashes again. Today this is masked only because polling stops entirely (finding 1). Any refetch trigger (the `['remote']` invalidation from mutations, `refetchOnReconnect` on OS online events, or anyone later adding backoff or a refresh button) resurrects the exact symptom from https://github.com/nebari-dev/nebi/issues/217. Suggest gating the spinner on first load only (e.g. `errorUpdateCount === 0`) rather than `!remoteError`.

**3. `pollUnlessErrored` on the `useRemoteServer` heartbeat removes the app's only liveness self-heal** (`frontend/src/hooks/useRemote.ts`)
That endpoint is a loopback DB read; it never errors due to remote unreachability. But one transient local-backend hiccup (e.g. desktop backend restart) permanently stops the 10s status poll, and its observer lives in `Layout.tsx`, which never unmounts. Connection status then freezes app-wide: remote view silently degrades to local-only with no banner, and Settings can show the connect form against an already-connected server. This hook should keep the plain interval.

**4. The `fetchMode` catch branch can recreate the wedge this PR fixes** (`frontend/src/store/modeStore.ts`)
The Wails app starts the embedded server asynchronously, so the single `/version` call can race it and fail. The catch then pins `mode: 'team'` and `networkMode: 'online'` for the whole session, so a desktop session that hits the race plus an OS offline report wedges every loopback query in `paused`, which is exactly the https://github.com/nebari-dev/nebi/issues/217 scenario. Suggest retrying `/version` before falling back, or not downgrading `networkMode` when the platform is known to be desktop.

### Non-blocking

- **AdminDashboard has the same gap this PR fixes elsewhere**: it consumes `useRemoteWorkspaces`/`useRemoteJobs` but got no banner or error handling, so with the remote down it renders frozen "0 workspaces / 0 jobs" tiles indefinitely. The `isRemoteConnected && viewMode === 'remote' && remoteError` derivation plus spinner tweak is now copy-pasted byte-identically in two places and missed a third; a shared hook (e.g. `useRemoteUnreachable()` or a richer `useRemoteView()`) would make the guarantee structural instead of per-page.
- **Banner message overclaims**: the backend wraps every remote failure as 502 (`Remote error: %v`), including remote-side 401s (expired token) and 500s, so "Remote server unreachable" can be wrong while the stopped polling makes it stick.
- **Transient banner against a healthy server**: disabled queries are skipped by `invalidateQueries`' refetch, so after disconnecting an unreachable server A and connecting healthy server B, the stale `isError` (data retained, so status is not reset to pending) shows the banner against B until the first fetch resolves.
- **Scope note**: the app-wide `networkMode` default change (every query and mutation, including team mode now pausing while offline) is a second, much broader fix riding under a banner-fix title. Worth a mention in the PR description so it gets changelog visibility, since every future query silently inherits it. Relatedly, two invariants this PR relies on are worth documenting in AGENTS.md: the networkMode policy per mode, and the fact that `/remote/server`'s `connected` means "URL + token stored", not "remote is reachable". If that status ever becomes a real liveness probe, the banner would hide exactly when it should show.
- `modeStore.test.ts` `beforeEach` re-hardcodes the queryClient constructor defaults, so the "pre-existing defaults are preserved" assertion will silently drift from `lib/queryClient.ts`; snapshotting `getDefaultOptions()` at module load avoids the second copy.
- `alert.tsx`: `AlertTitle` and the `destructive` variant have zero call sites, and the repo already trims shadcn components to used parts (`card.tsx`). Suggest dropping them.

jbouder:
Thanks for the thorough review, @tylerpotts — all four blocking findings and the non-blocking items are addressed in 4768d96.

### Blocking

1. **Polling never recovers after an error** — `pollUnlessErrored` is now `pollWithErrorBackoff`: on error it returns a 30s interval (`ERROR_BACKOFF_INTERVAL`) instead of `false`, so polling self-heals once the server comes back and the banner clears on its own.
2. **Dead `!remoteError` spinner guard** — the full-page spinners in `Jobs.tsx`, `Workspaces.tsx` (and now `AdminDashboard.tsx`) gate on `isLoading && errorUpdateCount === 0`, i.e. first load only. Refetches after an error happen behind the banner instead of resurrecting the #217 flash.
3. **`useRemoteServer` heartbeat** — reverted to a plain 10s `refetchInterval`, with a comment explaining why it must never back off (loopback DB read, only liveness self-heal, observer in `Layout` never unmounts).
4. **`fetchMode` catch branch** — `/version` is now retried (3 attempts, 300ms apart) to cover the embedded-server startup race, and on final failure `networkMode` is only downgraded to `'online'` when the Wails runtime is absent — the desktop app keeps `'always'`, so the #217 wedge can't be recreated. Covered by two new tests.

### Non-blocking

- **Shared derivation** — added `useRemoteView()` in `useRemote.ts` returning `{ isLocalMode, viewMode, isRemoteConnected, isRemoteView }`; Jobs, Workspaces, and AdminDashboard now use it, and AdminDashboard gets the unreachable banner plus error-aware loading like the other two pages.
- **Banner overclaim** — copy now covers an invalid session as well as a down server, with a comment noting the backend wraps remote-side auth failures as 502.
- **Stale banner against a healthy server** — connect/disconnect mutations now `resetQueries(['remote'])` instead of invalidating, so errored state from an unreachable server A can't survive into a fresh connection to server B (reset clears disabled queries too, which invalidation skips).
- **Scope note** — PR description updated to call out the app-wide `networkMode` change, and both invariants (networkMode policy per mode; `/remote/server`'s `connected` = "URL + token stored", not liveness) are documented in AGENTS.md.
- **`modeStore.test.ts` drift** — the beforeEach now restores defaults snapshotted from `queryClient.getDefaultOptions()` at module load instead of re-hardcoding them.
- **`alert.tsx`** — trimmed `AlertTitle` and the `destructive` variant (matching the `card.tsx` precedent).

Full suite passes: 199/199 tests, typecheck, Biome, and production build clean.

tylerpotts (changes requested):
Thanks for this fix, the banner + self-healing polling approach is the right shape, and the AGENTS.md additions documenting the new invariants are appreciated. One blocking concern and a few smaller points.

### Blocking

**The `useRemoteView()` invariant this PR documents isn't actually true yet.** AGENTS.md now says `useRemoteView()` is "the shared derivation pages should use to gate remote data and the unreachable banner", but only 3 of the 8 call sites are converted. The old hand-rolled derivation (`isLocalMode && serverStatus?.status === 'connected'` + `viewMode`) still lives in:

- `frontend/src/pages/Registries.tsx`
- `frontend/src/pages/admin/UserManagement.tsx`
- `frontend/src/pages/admin/AuditLogs.tsx`
- `frontend/src/pages/admin/RegistryManagement.tsx`
- `frontend/src/components/layout/Layout.tsx` (this one may be fine as-is since it's the self-heal observer)

Beyond doc drift, this is user-visible: Workspaces, Jobs, and the Admin Dashboard now show the unreachable banner, while Registries, User Management, and Audit Logs render remote-view pages with no banner when the same server is down. Either convert the remaining pages here, or soften the AGENTS.md wording to "new/updated pages should use" and open a follow-up issue so the doc matches reality.

### Questions / suggestions

1. **AdminDashboard's banner can stick after the server recovers.** `useRemoteDashboardStats` has no `refetchInterval`, but it's part of the `remoteUnreachable` OR on the dashboard. With `refetchOnWindowFocus: false`, an errored stats query only refetches on remount, so the banner self-heals for workspaces/jobs but not for stats until the user navigates away. Consider giving the stats query `pollWithErrorBackoff` too, or dropping it from the banner condition.

2. **The gating logic got duplicated one layer above the abstraction built for it.** `remoteUnreachable = isRemoteView && isError` and the `isLoading && errorUpdateCount === 0` first-load gate are copy-pasted (same comment included) into Workspaces, Jobs, and AdminDashboard. `errorUpdateCount` is a TanStack Query internal that's now leaked into three pages. Could the remote hooks (or `useRemoteView`) return `isFirstLoad` / `isUnreachable` so pages consume a named concept instead of re-deriving it?

3. **Scope note:** the `networkMode` default change is called out in the PR body (thanks), but the 3-attempt/300ms retry loop around `/version` and the desktop-detection fallback are also bootstrap behavior changes riding along in a banner bugfix. Fine with keeping them if intentional, just flagging that they delay team-mode fallback by up to ~600ms at startup and deserve a mention in the PR description.

4. **Two `/version` fetch paths now exist with different semantics:** `modeStore.fetchMode` uses raw `apiClient` with its own new retry policy, while `useVersion.ts` wraps the same endpoint in react-query. If bootstrap ordering forces the manual fetch (queryClient defaults must be set before queries run), a comment in `modeStore.ts` saying so would stop someone from "fixing" it later.

### Nits

- `modeStore.ts` re-implements the Wails detection idiom that already lives in `lib/openExternal.ts` (and silently depends on the `Window.runtime` type declared there). A shared `isDesktopApp()` helper in `lib/` would tidy this.
- `pollWithErrorBackoff` types its argument as a hand-rolled `{ state: { status: string } }`; TanStack's `Query` type would be tighter for an exported utility.
- `alert.tsx` following the trimmed-primitive pattern from `card.tsx` looks good. If a second colored alert shows up later, promote the banner's className overrides to a proper `variant`.

jbouder:
Thanks @tylerpotts — the blocking item is addressed in a6bff95, taking the second of your two options:

- **AGENTS.md wording softened** to "new or updated pages should gate remote data and the unreachable banner with `useRemoteView()`", and it now explicitly names which pages use it today versus which still hand-roll the derivation, so the doc matches reality.
- **Follow-up issue opened:** #504 tracks migrating `Registries`, `UserManagement`, `AuditLogs`, and `RegistryManagement` (with `Layout` to be either migrated or annotated as the intentional direct subscription). It also folds in your suggestion 2 — moving the `errorUpdateCount` first-load gate behind named concepts like `isFirstLoad`/`isUnreachable` on the hooks — since that's the same refactor.

On the remaining points, deliberately not changed in this push to keep the PR from growing further:

1. **AdminDashboard stats banner stickiness** — good catch; noted as a real edge (errored stats query only refetches on remount). Happy to add `pollWithErrorBackoff` to `useRemoteDashboardStats` here if you'd prefer it in this PR, otherwise it can ride with #504's rework of the banner condition.
2. Folded into #504 as above.
3. **Scope note** — agreed the `/version` retry (up to ~600ms added before team-mode fallback) is bootstrap behavior riding along; I'll add a line to the PR description.
4. The two `/version` paths / `isDesktopApp()` helper / `Query` type nits — all reasonable; leaving them for a cleanup pass unless you want them in this PR.

jbouder:
Took you up on question 1 after all: 1799ca6 gives `useRemoteDashboardStats` `pollWithErrorBackoff(30000)`, so AdminDashboard's banner now self-heals for all three of its queries instead of sticking until remount when only the stats query was errored. 30s rather than the 5s the list queries use, since dashboard tiles don't need tighter freshness. Suite still 199/199, typecheck and Biome clean.

tylerpotts (approved):
Approving. The latest commits resolve everything I flagged as blocking:

- **AGENTS.md wording** now matches reality: it names exactly which surfaces use `useRemoteView()` today, and the migration of the remaining hand-rolled call sites is tracked in https://github.com/nebari-dev/nebi/issues/504, which also absorbs the `errorUpdateCount`-in-pages cleanup with clear acceptance criteria. This is exactly the resolution I suggested.
- **Dashboard stats polling**: `useRemoteDashboardStats` now uses `pollWithErrorBackoff`, so all three queries feeding the AdminDashboard banner self-heal after the server recovers.
- **Scope note** in the PR body now discloses the `networkMode` default change, the `/version` retry loop, and the desktop fallback. Good for changelog visibility.

Two non-blocking leftovers, fine to land as-is or fold into follow-ups:

1. `modeStore.fetchMode` still bypasses the react-query `/version` path (`useVersion.ts`) with its own retry semantics and nothing says why. A one-line comment noting the bootstrap-ordering reason would stop someone from "unifying" it later.
2. The Wails detection in `modeStore.ts` duplicates the idiom in `lib/openExternal.ts` and leans on the `Window.runtime` type declared there. A shared `isDesktopApp()` in `lib/` would tidy this whenever someone is next in the area.

One observation, no action needed: `pollWithErrorBackoff(30000)` on the stats query is currently a no-op since the error backoff interval is also 30s. Using the wrapper anyway keeps the pattern uniform and the comment explains the intent, so it reads fine.

PR #505, opened 2026-08-13
feat: adopt the registry theme hook and align the profile menu

jbouder:
## Summary

nebi-scoped subset of nebari-dev/nebari-design#131: adopt the registry-managed theme hook/provider and align the profile dropdown menu with the shared header design. **The rest of the header (logo, nav links, view-mode toggle, admin button) is intentionally untouched.**

## Theme hook / provider

- Replaced the local `useThemePreference` + `useLocalStorageState` with the registry `@nebari/use-theme-preference` files (`src/hooks/use-theme-preference.ts` + `src/hooks/theme-provider.tsx`). They're vendored verbatim from the registry source rather than installed via `shadcn add`, because the registry item's `@nebari/theme` dependency would inject the nebari token set into nebi's own theme CSS.
- App wrapped in the registry `ThemeProvider` in `main.tsx`; `App.tsx` reads theme state via `useTheme()` and keeps the existing prop flow to `Login`/`Layout`.
- Existing storage key preserved via `storageKey={THEME_MODE_STORAGE_KEY}` (`nebi:themeMode`, now in `src/lib/theme.ts`) — users keep their saved preference.
- `index.html` gains the pre-paint bootstrap script (exact `themeBootstrapScript("nebi:themeMode")` output), so a saved dark preference no longer flashes light while the bundle loads.

## Profile menu

Restyled to match the canonical profile menu (reference: nebari-landing `Header.tsx`), keeping nebi's existing hand-rolled dropdown mechanics and trigger:

- Menu width 248px (`w-[248px] p-2`) with a name/email section at the top
- Theme picker stays a `menuitemradio` radio group, restyled as the shared segmented control (`h-[34px]` track, bordered/carded selected state)
- Sign-out item uses a new `--color-sign-out-foreground` token (`#d2161c` light / `#ff6b6b` dark) instead of hardcoded `text-red-500`

## Tests / checks

- `npm run check` (biome) — clean apart from one `useComponentExportOnlyModules` warning on the vendored registry `theme-provider.tsx` (same as in the other #131 adoptions) and two pre-existing biome-config migration infos
- `npm test` — 29 files / 200 tests pass; the theme-hook tests were migrated to the registry hook (including a new stored-preference read test), and the existing ProfileMenu suite (menuitemradio + `aria-checked`, keyboard nav, outside-click/Escape) passes against the new styling
- `npm run build` (`tsc -b` + vite) — pass (pre-existing >500 kB chunk warning)

- [ ] Attach light/dark profile-menu screenshots

Part of nebari-dev/nebari-design#131

https://claude.ai/code/session_012Xb4VPMBJ4ksnCiChJpVzu


jbouder:
Closing, need to migrate to NDS first.

PR #511, opened 2026-08-17
chore: auto-open a PR when the pinned Go toolchain goes stale

tylerpotts:
## Summary

Stops the problem behind https://github.com/nebari-dev/nebi/pull/510 from needing a human every time. Every Go patch release closes standard library vulnerabilities, so the day one ships, the govulncheck gates and the Trivy image scan begin failing on `main` and on **every open PR at once**. That is exactly how `go1.26.5` surfaced: as an unrelated frontend PR going red, not as a toolchain signal. `check-go-toolchain.sh` only warns on a stale patch outside the release gate, so nothing opened the bump.

This adds a daily job that turns that into a two-line PR waiting for review.

## Changes

- **`scripts/bump-go-toolchain.sh`** rewrites both places the toolchain is pinned: the `toolchain` directive in `go.mod` and the `golang` builder image tag + digest in the `Dockerfile`. Follows the existing `check-go-toolchain.sh` idiom (bash + `curl` + `python3`, no new dependencies) and is runnable locally.
- **`.github/workflows/go-toolchain-bump.yml`** runs it daily plus on `workflow_dispatch`, and opens the PR.

## Design decisions worth reviewing

**The digest is verified, not trusted.** After resolving the multi-arch index digest, the script pulls the image config and confirms it reports the expected `GOLANG_VERSION` before pinning. Trusting the tag alone would let a mismatched or mid-push image get pinned by digest, which is the one failure mode a digest pin exists to prevent. This is the same check I ran by hand for #510.

**Patch bumps only.** A major-line move also decides the language version and the `go` directive, so an unsupported line exits 2 and fails the run loudly instead of opening a PR.

**Idempotent via a version-keyed branch** (`chore/go-toolchain-<version>`), so a run that finds the previous bump PR still open leaves it alone.

## One caveat that needs a decision

A PR opened with the default `GITHUB_TOKEN` **does not trigger other workflows**, so the bump PR would sit with no checks. The workflow uses `secrets.BOT_TOKEN || secrets.GITHUB_TOKEN`, so:

- Set a `BOT_TOKEN` secret (PAT or GitHub App token) and bump PRs get full CI.
- Without it, the PR still opens correctly but needs a close/reopen to run checks. The generated PR body says so explicitly.

Creating PRs with `GITHUB_TOKEN` also requires *Settings -> Actions -> Allow GitHub Actions to create and approve pull requests* to be enabled.

## Test plan

Verified locally before pushing:

- [x] **Reproduces the real fix byte-for-byte.** Reset `go.mod` and `Dockerfile` to the pre-fix `go1.26.5` state and ran the script: the resulting diff against #510's known-good commit is **empty**.
- [x] Already-current path: exits 0, `changed=false`, no writes.
- [x] Unsupported line (`go1.20.0`): exits 2 with the manual-upgrade error.
- [x] Missing `toolchain` directive: exits 1.
- [x] Workflow YAML parses; the PR step's shell passes `bash -n`.
- [x] Dry-ran the PR step with `git`/`gh` stubbed: commit message and PR body render correctly (variables expanded, backticks literal, no stray indentation) and the `git`/`gh` invocations are right.
- [x] Idempotency path: with the branch already present upstream, the step short-circuits without pushing.

Note that CI here exercises the repo's normal gates, not this workflow, which only runs on its schedule. The verification above is where the behavior was checked.


.github/workflows/go-toolchain-bump.yml
```diff
@@ -0,0 +1,122 @@
...
+          DIGEST: ${{ steps.bump.outputs.digest }}
+        run: |
+          set -euo pipefail
+          BRANCH="chore/go-toolchain-${VERSION}"
+
+          # Version-keyed branch name doubles as the idempotency check: a run
+          # that finds last run's PR still open has nothing to add.
+          if git ls-remote --exit-code --heads origin "$BRANCH" >/dev/null 2>&1; then
+            echo "$BRANCH already exists upstream; leaving the open PR alone"
+            exit 0
```
MUFFANUJ, line 65: This idempotency check treats “branch exists” as the same thing as “there is already an open bump PR,” but those states can diverge. If a previous run successfully pushed `chore/go-toolchain-${VERSION}` and then failed during `gh pr create`, or if the PR was closed while the branch remained, every future run will hit this branch check and exit 0, leaving the stale Go pin without an open PR. Maybe we can check for an open PR for this head branch instead, or handle the “branch exists but no open PR” case by creating/reopening the PR or failing loudly?

scripts/bump-go-toolchain.sh
```diff
@@ -0,0 +1,146 @@
...
+digest = resp.headers["Docker-Content-Digest"]
+
+# Verify via one concrete platform: the index itself carries no version label.
+amd64 = next(
+    m["digest"] for m in index["manifests"]
+    if m.get("platform", {}).get("architecture") == "amd64"
+    and m.get("platform", {}).get("os") == "linux"
+)
+config = json.load(fetch(f"{registry}/manifests/{amd64}", token, ACCEPT))["config"]["digest"]
+env = json.load(fetch(f"{registry}/blobs/{config}", token))["config"].get("Env", [])
```
MUFFANUJ, line 106: This only verifies the `linux/amd64` child manifest before pinning the multi-arch digest, but the Docker workflow also builds `linux/arm64`. If Docker Hub had a bad or mid-push arm64 manifest while amd64 was correct, this script would still pin the manifest-list digest and the arm64 CI image would use an unverified builder. Should we verify the config for every platform Nebi builds, at least `linux/amd64` and `linux/arm64`, before accepting the digest?

PR #516, opened 2026-08-18
Support a dark-mode branding logo via logoUrlDark

jbouder:
## Summary

Nebi's runtime branding config (`/public/config.json`) accepted only a single `logoUrl`, and that one asset was used in both color schemes once an operator set it. The `isDarkMode` plumbing already existed — `getBrandingLogoUrl(isDarkMode)` is called from `Layout.tsx` and `Login.tsx` — but the flag only picked between the two *built-in* Nebi logos, so a custom logo could never vary by theme.

This adds `branding.logoUrlDark`, following other nebari software pack fallback semantics since Nebi already matches its structure.

## Changes

- **`frontend/src/lib/brandingConfig.ts`**
  - `logoUrlDark?: string` on the `BrandingConfig` type.
  - `sanitizeBrandingConfig` normalizes it and validates it through the existing `isSafeAssetUrl` gate (same-origin assets and allow-listed base64 `data:` image URIs only). It also now counts toward the "is there any branding at all?" check, so a config setting *only* a dark logo isn't discarded.
  - `getBrandingLogoUrl` resolves dark mode as `logoUrlDark ?? logoUrl`, light mode as `logoUrl`, then falls back to the built-in per-scheme default.
- **`frontend/public/public/config.json`** — seeds `"logoUrlDark": ""` after `logoUrl`.
- **`frontend/src/lib/brandingConfig.test.ts`** — 5 new cases.

## Resolution behavior

| `logoUrl` | `logoUrlDark` | Dark mode shows |
|---|---|---|
| `/acme.svg` | *(empty)* | `/acme.svg` |
| `/acme.svg` | `/acme-dark.svg` | `/acme-dark.svg` |
| *(empty)* | *(empty)* | `/nebi-logo-dark.svg` (built-in) |
| *(empty)* | `/acme-dark.svg` | `/acme-dark.svg` |

Existing deployments are unaffected: an operator with a single `logoUrl` keeps getting it in both themes, exactly as before.

Empty strings are treated as absent rather than as values — `normalizeString` maps `""` to `undefined` before the fallback runs — so the checked-in `"logoUrlDark": ""` falls through cleanly instead of resolving to an empty ``.

## Notes

- No Go changes needed: `internal/api/router.go` serves the runtime config file verbatim and never parses branding fields.
- The other two `config.json` copies (`frontend/dist/`, `internal/web/dist/`) are gitignored build artifacts.
- Nebi keeps its nested `branding` shape, so it stays consistent with chat-pack but still differs from landing / LLM serving / apps-pack, which put these keys at the top level. Any shared operator-facing branding docs will need to describe both layouts.

## Testing

- `npx tsc --noEmit` — clean
- `npx biome check` — clean
- `npx vitest run src/lib/brandingConfig.test.ts` — 14/14 passing


aktech (approved):
Looks good!

PR #517, opened 2026-08-21
fix: remove max_packages cap breaking registry imports

### LINKED ISSUE #502: [BUG] -  Pixi.toml not available for workspaces pulled from registry
### Description

The workspaces that I pull from OCI registries don't seem to have the pixi.toml stored or I can't access it. 

### Steps to reproduce

1. Pull a workspace from a registry
2. Let it build
3. Go to the workspace
4. Click configuration
5. Witness that the pixi.toml fails to load

Image

### Expected behavior

_No response_

### Environment

_No response_

### Additional context

_No response_

aktech:
Importing any environment with more than 128 transitive packages (every repo in the default registry qualifies) failed after a successful build: snapshot validation applied `max_packages` to the `pixi list` output, the create job failed, cleanup deleted the workspace directory, and the UI showed "Failed to load pixi.toml".

(This limit was added in https://github.com/nebari-dev/nebi/pull/494 and it does not makes sense to limit by number of packages)

- First commit stops applying the cap to resolver output.
- Second commit removes `max_packages` / `NEBI_LIMITS_MAX_PACKAGES` entirely (breaking: the knob is gone, ignored if still set). Abuse stays bounded by `manifest_bytes`, `package_string_bytes`, job quotas, and job timeout/CPU/storage budgets.

Fixes #502


internal/service/workspace_packages_test.go
```diff
@@ -106,60 +106,26 @@ func TestInstallPackages_NotFound(t *testing.T) {
 	}
 }
 
-func TestInstallPackages_RejectsTooManyPackagesBeforeJobWrite(t *testing.T) {
+func TestInstallPackages_AllowsLargePackageRequest(t *testing.T) {
+	// There is no cap on how many packages a request may name; abuse is
```
pmeier: I mean this comment makes sense historically, but it won't make much sense for someone reading this never knowing there ever was a limit.

internal/service/workspace_test.go
```diff
@@ -224,35 +225,26 @@ func TestCreate_RejectsOversizedPixiTomlBeforeWrites(t *testing.T) {
 	}
 }
 
-func TestCreate_RejectsTooManyManifestPackagesBeforeWrites(t *testing.T) {
+func TestCreate_AllowsManifestWithManyPackages(t *testing.T) {
+	// Manifest package count is not capped; manifest size is bounded by
```
pmeier: Same as above

internal/service/workspace_worker_test.go
```diff
@@ -293,46 +293,13 @@ func TestCreateVersionSnapshot_RejectsOversizedLockBeforeVersionWrite(t *testing
...
-	var versions int64
-	db.Model(&models.WorkspaceVersion{}).Where("workspace_id = ?", ws.ID).Count(&versions)
-	if versions != 0 {
-		t.Fatalf("expected no version writes, got %d", versions)
-	}
-}
-
-func TestCreateVersionSnapshot_RejectsTooManyListedPackagesBeforeVersionWrite(t *testing.T) {
+func TestCreateVersionSnapshot_StoresAllResolvedPackages(t *testing.T) {
+	// Resolver output is never count-capped: a small manifest routinely
```
pmeier: Same. Last flag for this. Applies to all similar places throughout this PR.

pmeier (approved):

internal/service/workspace_packages_test.go
aktech: Yes, makes sense!

PR #520, opened 2026-08-24
ci: publish multi-arch images (linux/amd64 + linux/arm64)

### LINKED ISSUE #417: Publish multi-arch container images (linux/amd64 + linux/arm64) to quay.io/nebari/nebi
## Summary

Container images published to `quay.io/nebari/nebi` are amd64-only. On a Kubernetes cluster with a mix of architectures (e.g. an amd64 general node group and an arm64 node group for storage, GPU, or batch workloads — increasingly common on AWS Graviton, GCP Tau T2A, and bare-metal arm), the nebi pod can land on an arm64 node and immediately fail:

```console
$ kubectl describe pod nebi-pack-... | grep -A2 Failed
  Warning  Failed   …  kubelet  Failed to pull image "quay.io/nebari/nebi:sha-be3187c":
    rpc error: code = NotFound desc = failed to pull and unpack image
    "quay.io/nebari/nebi:sha-be3187c": no match for platform in manifest: not found
  Warning  Failed   …  kubelet  Error: ErrImagePull
  Warning  Failed   …  kubelet  Error: ImagePullBackOff
```

## Confirming the platform set

The image is published as an OCI image index, but the index only contains a runnable amd64 entry plus a non-runnable attestation entry:

```console
$ curl -sS \
    -H 'Accept: application/vnd.oci.image.index.v1+json' \
    https://quay.io/v2/nebari/nebi/manifests/sha-be3187c | jq '.manifests[].platform'
{ "os": "linux", "architecture": "amd64" }
{ "os": "unknown", "architecture": "unknown" }
```

(The `unknown/unknown` entry is typical for SBOM/SLSA attestation manifests written by `docker buildx` — it's not a runnable image and Kubernetes correctly ignores it.)

So functionally the image is single-arch.

## Why this matters

Without an arm64 manifest, every deployment of nebi has to wrap its Pod spec in a `kubernetes.io/arch=amd64` nodeSelector — or trust the scheduler not to pick an arm node. The chart at `nebari-nebi-pack` doesn't currently expose a values-path for `nodeSelector` (separate issue: `nebari-dev/nebari-nebi-pack#35`), so the workaround today is a live `kubectl patch`. That live patch is lost on every cluster bootstrap and silent if anyone deploys nebi via raw manifests without knowing the constraint.

Multi-arch images solve this at the source: Kubernetes' image-platform matching at pull time uses the manifest list to filter candidate nodes, so a pod with no nodeSelector still goes to a node that can run the image. No chart change required, no nodeSelector required, no runtime patching.

## Proposed change

Extend the existing image build pipeline to produce both `linux/amd64` and `linux/arm64` variants and publish a multi-arch manifest list under the same tag.

Concretely (in whatever CI builds the image — GitHub Actions / GitLab CI / etc.):

```yaml
# example with docker/build-push-action -- adjust to actual CI shape
- uses: docker/setup-qemu-action@v3
- uses: docker/setup-buildx-action@v3
- uses: docker/build-push-action@v6
  with:
    context: .
    platforms: linux/amd64,linux/arm64
    push: true
    tags: quay.io/nebari/nebi:sha-${{ github.sha }}
    # plus :latest, :semver tags as already done
```

Same change applies to any sibling images published by this repo (CLI, desktop, etc., if applicable).

## Acceptance criteria

- [ ] `quay.io/nebari/nebi:<any-released-tag>` manifest list contains both `linux/amd64` and `linux/arm64` runnable entries.
- [ ] Pulling the image on an arm64 node succeeds without specifying `--platform`.
- [ ] Existing amd64 deployments are unaffected (same digest, same behavior on amd64 nodes).
- [ ] CI publishes both arches for every tag going forward.

## Out of scope

- **Native arm64 binaries for non-container distributions** (pixi global install, .deb packages, etc.) — useful but separate. This issue covers the container image only.
- **Testing arm64 functionality.** Just publishing the manifest is enough to unblock the scheduling problem; functional arm64 testing can come later if specific arm-only bugs surface. (Most Go applications run cleanly on arm64 without code changes; nebi is written in Go.)

## Related

- `nebari-dev/nebari-nebi-pack#35` — chart-side workaround needed until this image issue lands.
- `nebari-dev/nebari-nebi-pack#20` — separate AuthConfig fields issue, surfaced during the same dev-cluster validation.

aktech:
Images on quay.io were amd64-only, so pods scheduled to arm64 nodes fail with ImagePullBackOff. CI now builds linux/amd64 and linux/arm64 natively (ubuntu-latest and ubuntu-24.04-arm), smoke-tests and scans each platform on its own runner, pushes by digest, and merges the digests into one manifest list under the same tags.

Closes #417


Dockerfile
```diff
@@ -29,9 +29,13 @@ COPY --from=frontend-builder /app/frontend/dist ./internal/web/dist
...
 RUN make swagger
 
-# Build pure Go binary with CGO disabled
+# Build pure Go binary with CGO disabled. TARGETOS/TARGETARCH are set by
+# BuildKit to the build's target platform; CI builds each platform on a
+# native runner, so this is always a native compile there.
 ARG VERSION=dev
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 \
+ARG TARGETOS
+ARG TARGETARCH
```
pmeier, line 37: Any chance we can set defaults here for builds that happen outside of the CI, e.g. local dev?

pmeier (approved):

Dockerfile
aktech: These arguments are automatically set by buildkit: https://docs.docker.com/reference/dockerfile/#automatic-platform-args-in-the-global-scope

setting default would mean breaking the build when running docker build on non-default platform

PR #521, opened 2026-08-24
Allow failed workspaces to accept jobs so they can recover (#497)

### LINKED ISSUE #497: [BUG] A failed solve leaves a workspace permanently in `failed` with no recovery path
If a solve fails, the workspace goes to `status: "failed"` and stays there for good. The solve endpoint won't run again on a failed workspace, so even after you correct the spec to something perfectly valid, you can't re-solve. There's no reset/retry/rebuild in the API, nothing in the CLI, and as far as I can tell nothing in the UI either. The only ways out I found are editing `nebi.db` by hand or deleting and recreating the workspace.

This is really easy to hit. A single typo in a version pin bricks the workspace with no in-product way back.

I reproduced it end-to-end against a throwaway managed workspace via the local nebi API (`/user/<user>/nebi/api/v1`). Same flow from the UI: create a managed workspace, give it an unsolvable dependency, Solve, then fix the dependency and Solve again.

```
# fresh workspace
initial status: ready
scaffolded files: ['.gitattributes', '.gitignore', 'pixi.lock', 'pixi.toml']

# STEP 1 - unsolvable spec
wrote python="==3.99" -> POST /solve = 202 {"id":"eb0c80a9-...","workspace_id":"baf54442-..."}
   status=failed

# STEP 2 - valid spec, try to recover
wrote python="==3.13" -> POST /solve = 400 {"error":"Workspace is not ready"}
   status after recovery attempt: failed

# STEP 3 - look for any recovery endpoint
POST /workspaces/<id>/reset   -> 404
POST /workspaces/<id>/retry   -> 404
POST /workspaces/<id>/cancel  -> 404
POST /workspaces/<id>/rebuild -> 404
```

The CLI doesn't have a repair/reset command either. The subcommands are `init / shell / status / workspace / diff / import / publish / pull / push`.

I think a failed solve just shouldn't be terminal. Any of these would do it: let `POST /solve` run on a failed workspace and re-attempt from the current spec, or add an explicit reset/retry endpoint plus a UI button, or move a workspace back to a re-solvable state (`ready` / `needs-solve`) once its spec changes so the next Solve is allowed.

The `Workspace is not ready` message doesn't help either. It reads like a transient error, but it's actually a dead end, and it doesn't point at the spec or how to recover.

Current workaround is either editing `~/.local/share/nebi/nebi.db` to set the workspace `status` back to `ready` (risky, nebi is running with WAL active) then re-solving, or deleting the workspace and recreating it, which loses its env config.

This is adjacent to #432 (UI dep edits vs manual config edits) since both are about how spec edits and solve state interact, but this one is specifically the terminal `failed` state with no way back. Separately I noticed a plain out-of-band edit to `pixi.toml` on disk doesn't trigger a re-solve on its own, only an explicit Solve does. Might be worth its own issue if it isn't already tracked.

## Versions / environment

- nebi: `0.12.dev+be3187c` (commit `be3187c`, `go1.24.13`, `/api/v1/version` reports `mode: local`)
- nebi pack: `nebari-nebi-pack` chart `v0.1.0-alpha.7`, image tag `sha-be3187c`
- ds-pack jupyterlab image (nebi client binary): `quay.io/nebari/nebari-data-science-pack-jupyterlab:sha-46289ab`
- Deployment: NIC
- Cloud: Hetzner

tylerpotts:
Fixes https://github.com/nebari-dev/nebi/issues/497. Supersedes https://github.com/nebari-dev/nebi/pull/421.

## Problem

A failed solve sets the workspace status to `failed`, and `submitJob` rejected any workspace whose status was not `ready`. Nothing ever transitions `failed` back to `ready` except a successful job, and no job could be submitted, so a single bad version pin permanently bricked the workspace. There is no reset/retry endpoint, and the only workarounds were hand-editing `nebi.db` or deleting the workspace.

Reproduced end-to-end against a local-mode server built from current `main` (f7d95a9): unsolvable spec (`python = "==3.99"`) → `POST /solve` 202 → status `failed`; corrected spec → `POST /solve` 400 `{"error":"Workspace is not ready"}`, stuck in `failed` forever.

## Fix

Treat `failed` as job-eligible alongside `ready` in the two status gates:

- `submitJob` (`internal/service/workspace_packages.go`) — solve, install, remove, and rollback jobs can now be submitted on a `failed` workspace, so a corrected `pixi.toml` recovers it on the next solve.
- `RollbackToVersion` (`internal/service/workspace_worker.go`) — rolling back to a known-good version is also a valid recovery path from `failed`.

Transitional states (`pending`, `creating`, `deleting`) still reject, and the rejection message now includes the actual status (`Workspace is not ready (status: creating)`) instead of the misleading bare message the issue called out.

No new concurrency restrictions: admission is still governed by the existing active-job quotas (`checkActiveJobQuotas`), and while a solve is executing the workspace is in `creating`, which rejects as before. This is why the approach differs from #421 — see below.

## Why supersede #421

#421 predates the job-admission rework (admission transaction + `lockJobAdmission` + active-job quotas) and replaces the status gate with a `hasActiveJob` check that permits at most one active job per workspace. On current `main` that would hard-cap `ActiveJobsPerWorkspace` (configurable, default 2) at 1 and required rewriting several quota tests. The relaxed status gate fixes the actual bug (terminal `failed`) without adding that limitation.

## Testing

- New regression tests, written failing-first against the old gate: `TestSolveWorkspace_AllowedOnFailedWorkspace`, `TestRollbackToVersion_AllowedOnFailedWorkspace`.
- Full `go test ./internal/... ./cmd/...` passes with no changes to existing tests.
- Live verification against a local-mode server with real pixi 0.76.1: bad pin → `failed`; corrected pin → solve 202 → `ready` with a solved environment. A workspace bricked by the pre-fix binary in the same database was also recovered by a solve after upgrading, so already-stuck workspaces self-heal with no migration needed.


internal/service/workspace_packages.go
```diff
@@ -41,8 +41,11 @@ func (s *WorkspaceService) submitJob(ctx context.Context, wsID string, userID uu
 			return err
 		}
 
-		if ws.Status != models.WsStatusReady {
-			return &ValidationError{Message: "Workspace is not ready"}
+		// A failed workspace may accept new jobs so a corrected spec or a
+		// rollback can recover it (issue #497); only transitional states
+		// (pending/creating/deleting) reject job submission.
+		if ws.Status != models.WsStatusReady && ws.Status != models.WsStatusFailed {
```
pmeier: Instead of duplicating this check, can we add a small function to the type that also removes the need for a long comment

```go
func (w WorkspaceStatus) IsTransitional() bool {
	return !(w == WsStatusReady || w == WsStatusFailed)
}
```

internal/service/workspace_packages.go
```diff
@@ -41,8 +41,11 @@ func (s *WorkspaceService) submitJob(ctx context.Context, wsID string, userID uu
...
+			return &ValidationError{Message: fmt.Sprintf("Workspace is not ready (status: %s)", ws.Status)}
```
pmeier: "is not ready" is no longer true, is it? Maybe something like "is not ready to accept jobs"? Open to suggestions.

internal/service/workspace_worker.go
```diff
@@ -26,8 +26,9 @@ func (s *WorkspaceService) RollbackToVersion(ctx context.Context, wsID string, v
 		return nil, err
 	}
 
-	if ws.Status != models.WsStatusReady {
-		return nil, &ValidationError{Message: "Workspace is not ready"}
+	// Failed workspaces may roll back to a known-good version (issue #497).
+	if ws.Status != models.WsStatusReady && ws.Status != models.WsStatusFailed {
+		return nil, &ValidationError{Message: fmt.Sprintf("Workspace is not ready (status: %s)", ws.Status)}
```
pmeier: Same as above

internal/service/workspace_packages.go
tylerpotts: Fixed in adac519: added `WorkspaceStatus.IsTransitional()` in `internal/models/workspace.go` (with the rationale moved into its doc comment) and replaced both duplicated checks with it.

internal/service/workspace_worker.go
tylerpotts: Fixed in adac519: rollback now uses `ws.Status.IsTransitional()` as well.

internal/service/workspace_packages.go
tylerpotts: Updated in 2f5ba63 to make the placeholder explicit: `Workspace is not ready to accept jobs while status is: 'creating'`.

pmeier (approved):

PR #522, opened 2026-08-24
Remove uv and the package-manager abstraction; pixi is the only package manager

tylerpotts:
## Summary

uv was only ever stubbed out as a provider and never implemented. This removes it and the multi-package-manager abstraction that existed to support it. Pixi is the only package manager Nebi uses, and the extra indirection wasn't paying for itself.

### Abstraction removal
- Delete `internal/pkgmgr` (the `PackageManager` interface, factory, and registry). The pixi implementation moves up to `internal/pixi` and owns its option/result types directly.
- The executor and services call `*pixi.PixiManager` concretely. `internal/executor` no longer type-asserts to find the binary path.
- The old factory registry doubled as the test-injection seam. It's replaced by one narrow hook, `service.SetPixiListPackagesForTests`, which stubs `pixi list` — the only pixi dependency tests actually needed to fake (everything else goes through the injected `Executor`).

### uv removal
- `package_manager.default_type` and `package_manager.uv_path` config keys and their env bindings are gone. `package_manager.pixi_path` / `NEBI_PACKAGE_MANAGER_PIXI_PATH` still work unchanged.
- `docker/uv.Dockerfile` and the `build-docker-uv` Makefile target are deleted; `test-pkgmgr` is now `test-pixi`.
- `UV_CACHE_DIR` in `internal/process/env.go` intentionally stays — that's pixi's own uv cache for pypi-dependencies, not the provider.

### `package_manager` field removal (end to end)
The workspace field could only ever hold `"pixi"`, so it's dropped from the server model, the CLI local store, API requests/responses, the CLI client and `nebi info`, the Wails app, swagger (regenerated), and the frontend (types, create calls, and the Package Manager rows on both detail pages).

**Migration note:** the column was `NOT NULL` with no default, and GORM AutoMigrate never drops columns, so leaving it would break all workspace inserts on existing databases. Both `db.Migrate` (server) and `store.Open` (CLI local SQLite) now drop the column if present. Regression tests build the legacy schema and verify the drop plus a post-migration insert. Data loss is nil in practice — the column only ever held `"pixi"`.

**Compatibility:** old CLIs that send `package_manager` on workspace create are silently ignored; an old CLI running `nebi info` against a new server shows no package-manager line.

## Test plan
- `go build` / `go vet` clean, including `-tags=e2e`
- Full Go test suite passes (`./internal/... ./cmd/...`)
- New regression tests for the legacy-column migrations in `internal/db` and `internal/store`
- Frontend: `tsc -b` clean, all 233 vitest tests pass, biome CI check clean
- Swagger regenerated via `make swagger` (zero `package_manager` references remain)


internal/config/config.go
```diff
@@ -119,11 +119,9 @@ type LogConfig struct {
 	Level  string `mapstructure:"level"`  // "debug", "info", "warn", "error"
 }
 
-// PackageManagerConfig holds package manager configuration
+// PackageManagerConfig holds pixi configuration
 type PackageManagerConfig struct {
```
pmeier: Do we need to keep this struct given that it now only holds one field or can we use the field directly?

internal/executor/local.go
```diff
@@ -91,11 +90,10 @@ func (e *LocalExecutor) CreateWorkspace(ctx context.Context, ws *models.Workspac
 	return e.withStorageLimit(ctx, envPath, logWriter, func(ctx context.Context) error {
 		fmt.Fprintf(logWriter, "Creating environment at: %s\n", envPath)
 
-		pm, err := e.packageManagerFor(ctx, ws)
+		pm, err := e.pixiFor(ctx)
```
pmeier, line 93: Nit: this variable should probably just be `p` for pixi rather than `pm` for package manager


```suggestion
		p, err := e.pixiFor(ctx)
```

internal/pixi/types.go
```diff
@@ -78,7 +53,7 @@ type OutputLimitError struct {
 func (e *OutputLimitError) Error() string {
 	command := e.Command
 	if command == "" {
-		command = "package manager"
+		command = "pixi"
```
pmeier: How can `e.Command` be an empty string that we need to set a default here? Also if it actually can be empty for some reason, isn't setting a hardcoded value here misleading as the error was not produced by pixi in the first place?

internal/executor/local.go
pmeier: Well I see that the return is a pixi manager so `pm` is fine.

internal/config/config.go
tylerpotts: Fixed in 201ee7a: dropped `PackageManagerConfig` and made `PixiPath` a top-level field (`pixi_path` / `NEBI_PIXI_PATH`). `NEBI_PACKAGE_MANAGER_PIXI_PATH` is still bound as a legacy env alias so existing deployments keep working.

internal/pixi/types.go
tylerpotts: Agreed, it cannot be empty: every constructor sets `Command: "pixi list"` and a `Stream`. The fallbacks were a leftover from the generic package-manager version. Removed them in 201ee7a so `Error()` just formats the fields.

pmeier (approved):

PR #533, opened 2026-08-25
fix: hide the Admin menu and lock /admin routes in local mode

### LINKED ISSUE #524: Hide the Admin Menu (and Keep the Profile Menu Hidden) in Local Mode
### Summary

In local mode the top-right header still renders the `Admin` nav button when `useIsAdmin()` resolves true — but admin features (user management, groups, identity reviews, audit logs) are meaningless on a single-user local server where auth is bypassed. Gate the `Admin` button in `frontend/src/components/layout/Layout.tsx` on `!isLocalMode`, skip the `useIsAdmin()` probe entirely in local mode, and lock down the `/admin/*` routes so they aren't reachable by URL. The `ProfileMenu` is already hidden in local mode (`Layout.tsx:187`) — this issue adds a regression test to keep it that way.

### Motivation

- Local mode bypasses auth entirely (`App.tsx`: `if (isLocalMode) return children`), so there is no user or role for an admin dashboard to manage
- The `Admin` button (`Layout.tsx:174`) is gated only on `isAdmin`, not on mode, so it can appear in local mode
- `useIsAdmin()` determines admin status by firing a real `adminApi.getUsers()` request — a pointless round trip on every local-mode app load
- `/admin/*` routes stay navigable by direct URL in local mode even once the button is hidden
- Removes two header affordances that do nothing in local mode, leaving `Workspaces` / `Registries` / `Settings` / `Docs` as the actual surface

### Acceptance Criteria

- [ ] The `Admin` nav button in `Layout.tsx` renders only when `!isLocalMode && isAdmin`
- [ ] `useIsAdmin()` (`hooks/useAdmin.ts`) does not issue its `adminApi.getUsers()` probe in local mode — the query is disabled and resolves as not-admin
- [ ] `AdminRoute` in `App.tsx` redirects away from `/admin/*` in local mode instead of rendering the dashboard, and does not fire its own `is_admin` probe there
- [ ] `ProfileMenu` remains hidden in local mode, covered by a test so it can't regress
- [ ] Layout tests assert: no `Admin` button and no profile menu trigger in local mode; both present (given `isAdmin`) in team mode
- [ ] `pnpm test` / `biome check` pass with no new warnings

### Out of Scope

- Removing or restructuring the admin pages themselves (`pages/admin/*`) — they stay for team mode
- Server-side admin authorization — this is a UI visibility change only
- Adding a theme control for local mode. Hiding `ProfileMenu` also hides the theme picker, and `Settings.tsx` has no replacement; that gap should be its own issue if it matters
- The local-mode `Settings` nav item and the remote/local view-mode toggle — both are intentionally local-mode surfaces

jbouder:
## Summary

Local mode bypasses auth entirely, so the admin dashboard has nothing to manage there. This hides the admin surface in local mode:

- The `Admin` nav button in `Layout.tsx` renders only when `!isLocalMode && isAdmin`
- `useIsAdmin()` sets `enabled: !isLocalMode`, so the `adminApi.getUsers()` probe never fires in local mode and the hook resolves as not-admin — no wasted round trip on every local-mode app load
- `AdminRoute` redirects `/admin/*` to `/workspaces` in local mode, and now reuses `useIsAdmin()` instead of duplicating the inline `is_admin` query (so it inherits the local-mode disable)

## Tests

- New `Layout.test.tsx`: local mode shows neither the Admin button nor the profile-menu trigger, even with a stale user in the auth store (regression guard for the already-hidden `ProfileMenu`); team mode with an admin shows both
- New `App.test.tsx`: `AdminRoute` redirects in local mode without hitting the admin API (asserted via an msw spy), renders admin routes for team-mode admins, and redirects team-mode non-admins
- `useAdmin.test.ts`: the `is_admin` query stays idle and resolves as not-admin in local mode

Full suite passes (239 tests / 34 files), `tsc -b` clean, `biome check` clean with no new warnings.

Closes #524

### DIFF
--- a/frontend/src/App.test.tsx
+++ b/frontend/src/App.test.tsx
@@ -1,0 +1,63 @@
+import { HttpResponse, http } from 'msw';
+import { Route, Routes } from 'react-router-dom';
+import { afterEach, describe, expect, it } from 'vitest';
+import { AdminRoute } from './App';
+import { useModeStore } from './store/modeStore';
+import { server } from './test/handlers';
+import { renderWithProviders, screen } from './test/utils';
+
+const renderAdminRoutes = () =>
+  renderWithProviders(
+    <Routes>
+      <Route path="/workspaces" element={<div>workspaces page</div>} />
+      <Route element={<AdminRoute />}>
+        <Route path="/admin" element={<div>admin dashboard</div>} />
+      </Route>
+    </Routes>,
+    { initialEntries: ['/admin'] },
+  );
+
+describe('AdminRoute', () => {
+  afterEach(() => {
+    useModeStore.setState({ mode: null, features: {}, loading: true });
+  });
+
+  it('redirects away from /admin in local mode without probing the admin API', async () => {
+    useModeStore.setState({ mode: 'local', features: {}, loading: false });
+    let probed = false;
+    server.use(
+      http.get('/api/v1/admin/users', () => {
+        probed = true;
+        return HttpResponse.json([]);
+      }),
+    );
+
+    renderAdminRoutes();
+
+    expect(await screen.findByText('workspaces page')).toBeInTheDocument();
+    expect(screen.queryByText('admin dashboard')).not.toBeInTheDocument();
+    expect(probed).toBe(false);
+  });
+
+  it('renders admin routes for admins in team mode', async () => {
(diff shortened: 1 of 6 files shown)


jbouder:
Closed, going a different route

PR #466, opened 2026-07-26
Reject query-string auth tokens for protected routes

MUFFANUJ:
## Reference Issues or PRs
Closes #449 

## What does this implement/fix?

_Put a `x` in the boxes that apply_

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features not to work as expected)
- [ ] Documentation Update
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Other (please describe):

## Testing

- [x] Did you test the pull request locally?
- [x] Did you add new tests?

#### Text styling

- [ ] The content is written with [plain language](https://www.plainlanguage.gov/guidelines/) (where relevant).
- [ ] If there are headers, they use the proper header tags (with only one level-one header: `H1` or `#` in markdown).
- [ ] All links describe where they link to (for example, check the [Nebari website](https://nebari.dev/)).
- [ ] This content adheres to the Nebari style guides.

#### Non-text content

- [ ] All content is represented as text (for example, images need alt text, and videos need captions or descriptive transcripts).
- [ ] If there are emojis, there are not more than three in a row.
- [ ] Don't use [flashing GIFs or videos](https://www.w3.org/TR/UNDERSTANDING-WCAG20/seizure-does-not-violate.html).
- [ ] If the content were to be read as plain text, it still makes sense, and no information is missing.

## Any other comments?

Removes JWT authentication via `?token=` from protected routes, switches job log streaming to bearer-header `fetch()` streaming, and updates Swagger accordingly. Adds router/logging regression coverage and normalizes SSE log output with a shared multiline-safe writer.

PR #536, opened 2026-08-25
Make log broker overflow visible: bounded wait and drop marker

tylerpotts:
## Problem

The in-memory `LogBroker` does a non-blocking send to each subscriber and silently drops lines when the 100-entry channel buffer is full. pixi emits output in bursts, so the live job log in the UI had gaps that disappeared after a refresh (the DB copy was complete; only the stream lost data).

This is the second half of https://github.com/nebari-dev/nebi/pull/419. That PR's SSE framing fix (one `data:` prefix per line for multiline chunks) already landed via https://github.com/nebari-dev/nebi/pull/466; its remaining change was the buffer bump, which this PR includes and extends.

## Changes

- Per-subscriber buffer raised from 100 to 4096 lines.
- `Publish` stays non-blocking: on a full buffer the line is dropped for that subscriber. `Publish` runs synchronously in the job's stdout writer, and a subscriber 4096 lines behind is not worth stalling the job for.
- Drops are counted per subscriber. As soon as room frees up, a marker line `[nebi] N log lines dropped because the client could not keep up; refresh to see the full log` is delivered ahead of the next line, so a gap is visible rather than silent.
- Slow subscribers no longer cause drops for fast subscribers on the same job.
- Per-subscriber mutex so concurrent `Publish` calls keep marker/line ordering and the drop counter consistent.

## Tests

`internal/logstream/broker_test.go`:
- 1000-line burst delivered intact (fails at buffer=100 with 900 lost)
- drop on full buffer without blocking, marker emitted with correct count, dropped lines not delivered
- marker emitted exactly once per gap
- slow subscriber isolated from fast subscriber
- concurrent Publish under `-race`

`internal/api/handlers/job_stream_test.go`:
- browser-style SSE parser; end-to-end through `streamLogsFromBroker` with a concurrent producer of multiline chunks (incl. `\r\n`), asserting every line is `data:`-prefixed and none dropped. Mutation-checked: fails against both the old `data: %s\n\n` format and the old buffer size.

`go test ./internal/... -race` and `golangci-lint` are clean.

### DIFF
--- a/internal/logstream/broker.go
+++ b/internal/logstream/broker.go
@@ -3,7 +3,6 @@ package logstream
 import (
 	"fmt"
 	"sync"
-	"time"
 
 	"github.com/google/uuid"
 )
@@ -13,11 +12,6 @@ const (
 	// pixi emits output in bursts of many lines; a small buffer caused lines
 	// to be dropped before the SSE writer could drain them.
 	subscriberBufferSize = 4096
-
-	// defaultSendTimeout bounds how long Publish waits for a slow subscriber
-	// before dropping a line. Publish runs synchronously in the job's stdout
-	// writer, so this must stay short to avoid stalling the job itself.
-	defaultSendTimeout = 100 * time.Millisecond
 )
 
 // subscriber holds a subscriber's channel and how many lines have been
@@ -32,14 +26,12 @@ type subscriber struct {
 type LogBroker struct {
 	subscribers map[uuid.UUID]map[chan string]*subscriber // jobID -> set of subscribers
 	mu          sync.RWMutex
-	sendTimeout time.Duration
 }
 
 // NewBroker creates a new log broker
 func NewBroker() *LogBroker {
 	return &LogBroker{
 		subscribers: make(map[uuid.UUID]map[chan string]*subscriber),
-		sendTimeout: defaultSendTimeout,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


internal/logstream/broker.go
```diff
@@ -1,21 +1,45 @@
...
+const (
+	// subscriberBufferSize is the number of log lines buffered per subscriber.
+	// pixi emits output in bursts of many lines; a small buffer caused lines
+	// to be dropped before the SSE writer could drain them.
+	subscriberBufferSize = 4096
+
+	// defaultSendTimeout bounds how long Publish waits for a slow subscriber
+	// before dropping a line. Publish runs synchronously in the job's stdout
+	// writer, so this must stay short to avoid stalling the job itself.
+	defaultSendTimeout = 100 * time.Millisecond
```
pmeier: Not sure I agree with this being a thing. if we up our buffer to 4096, IIUC the only time we are going to wait is if the client is 4096 lines behind. Ignoring the fact that most logs likely will never reach this size in the first place, this means that the client is way to slow to handle streaming in the first place. I feel it would be ok to just drop lines in this case without any wait time. Especially if we include the number of dropped lines with the next message.

pmeier (approved):
One comment to resolve. Afterwards LGTM if CI is green.

internal/logstream/broker.go
tylerpotts: Agreed, removed the bounded wait in 51502a9. Publish is now strictly non-blocking: a full buffer drops the line for that subscriber and increments its counter, and the drop marker with the count is delivered ahead of the next line once there is room. sendTimeout and the timer path are gone; the bounded-wait test was removed and the drop test now asserts Publish returns within 100ms on a full buffer.

PR #538, opened 2026-08-26
Notify parent when Nebi jobs complete

MUFFANUJ:
## Reference Issues or PRs

Nebi part of the implementation from https://github.com/nebari-dev/data-science-pack/issues/227

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you test the pull request locally?
- [ ] Did you add new tests?

#### Text styling

#### Non-text content

## Any other comments?

Emits a same-origin `nebi:job-completed` message from the existing `useJobs` / `useJob` polling path when a refresh-relevant job transitions to completed. This lets an embedding parent such as JupyterLab Launchpad refresh kernels promptly without adding Jupyter-specific behavior to Nebi, while avoiding repeated messages on every poll.


frontend/src/hooks/useJobs.ts
```diff
@@ -1,19 +1,64 @@
...
 
 export const useJobs = () => {
-  return useQuery({
+  const query = useQuery({
     queryKey: ['jobs'],
     queryFn: jobsApi.list,
     refetchInterval: 2000, // Poll every 2 seconds for real-time updates
   });
+
+  useParentJobCompletionMessage(query.data);
```
tylerpotts: This is the change request. `useJobs()` has exactly two production callers, `components/jobs/Jobs.tsx:214` and `pages/admin/AdminDashboard.tsx:88`, and `useJob(` has none. `Jobs` is rendered in one place, inside `<TabsContent value="jobs">` at `pages/WorkspaceDetail.tsx:755-756`, and our `TabsContent` returns `null` for inactive tabs (`components/ui/tabs.tsx:80`). The default tab is `'overview'` (`WorkspaceDetail.tsx:79-81`), and `InstallControls` sits in the header (`:192`) and on the Workspaces list (`Workspaces.tsx:453`), which has no jobs hook at all. So: click Install from either place, don't open the Jobs tab, and no message is ever sent. Since `previousStatuses` starts empty per mount, opening the tab later doesn't recover it either.

Suggested shape, matching how the app already handles app-lifetime observers (`Layout.tsx:31-34` keeps `useRemoteView()` mounted for exactly this reason) and host detection (`isDesktopApp` in `store/modeStore.ts`):

- `src/lib/hostBridge.ts`: `isEmbedded()` (`window.parent !== window`), the message constant and a `NebiHostMessage` type, `postToHost(msg)` with `window.location.origin` as target.
- `src/hooks/useHostJobNotifications.ts`: owns a `useQuery({ queryKey: ['jobs'], queryFn: jobsApi.list, refetchInterval: 2000, enabled: isEmbedded() && isLocalMode })` plus the transition tracking, called once from `Layout.tsx` next to `useRemoteView()`. Same query key as `useJobs`, so no duplicate requests when the Jobs tab is also open, and non-embedded users don't pick up a permanent poll.
- `useJobs` / `useJob` go back to being pure wrappers.

frontend/src/hooks/useJobs.ts
```diff
@@ -1,19 +1,64 @@
 import { useQuery } from '@tanstack/react-query';
+import { useEffect, useRef } from 'react';
 import { jobsApi } from '@/api/jobs';
+import type { Job, JobStatus } from '@/types';
+
+const parentRefreshJobTypes = new Set<Job['type']>(['create', 'env_install']);
```
tylerpotts: Change request: this allowlist leaves Launchpad stale for the cases that matter most after install. `env_uninstall` removes `.pixi/envs` (`internal/executor/local.go:344-347`), so the kernel flips from Ready to not-installed and Launchpad keeps offering it for up to 61s. `delete` removes the workspace from `nebi workspace list`, which is what discovery reads. `install`/`remove`/`update`/`rollback` reinstall the env when it was installed and can add or drop `ipykernel`.

Simplest: drop the filter and post on every `completed` transition (Launchpad coalesces refreshes, and discovery is cheap). If you want to keep a filter, use the exhaustive `Record<JobType, boolean>` pattern from `components/jobs/Jobs.tsx:27-36` so a new job type forces a decision instead of silently opting out.

frontend/src/hooks/useJobs.ts
```diff
@@ -1,19 +1,64 @@
...
+  useEffect(() => {
+    const nextStatuses: Record<string, JobStatus> = {};
+    const jobList = Array.isArray(jobs) ? jobs : jobs ? [jobs] : [];
+
+    for (const job of jobList) {
+      const previousStatus = previousStatuses.current[job.id];
+      nextStatuses[job.id] = job.status;
+
+      if (
+        previousStatus !== undefined &&
```
tylerpotts: Non-blocking now, important after the observer is always mounted: `previousStatus !== undefined` means a job whose first observed status is already `completed` never posts. I checked with a throwaway test (first poll `[job-a running]`, second `[job-a running, job-b completed]`): no `postMessage`; the mirror assertion fails with "expected to be called 1 times, but got 0 times". Today the install mutation's `invalidateQueries(['jobs'])` refetches immediately, which mostly hides this, but a fast `create` can still slip through.

Suggest: treat the first successful fetch as initialization only, then count any job first seen as `completed` after that as a completion. A rare spurious refresh on list churn is harmless.

frontend/src/hooks/useJobs.ts
```diff
@@ -1,19 +1,64 @@
...
+        previousStatus !== 'completed' &&
+        job.status === 'completed' &&
+        parentRefreshJobTypes.has(job.type)
+      ) {
+        const jobType = job.type;
+        const workspace = job.workspace_id;
+
+        if (window.parent !== window) {
+          window.parent.postMessage(
+            { type: 'nebi:job-completed', jobType, workspace },
```
tylerpotts: This literal is now an external contract that Launchpad hardcodes, so changing it later is a breaking change. Worth an exported constant + type (see the `hostBridge.ts` suggestion above). While nothing on the receiving side reads the payload yet: `workspace` carries `workspace_id`, so `workspaceId` is the accurate name, and including `status` lets a host filter if it ever wants to.

`window.location.origin` as `targetOrigin` is the right default for the jupyter-server-proxy same-origin case; just worth a comment that cross-origin framing via `server.allowed_origins` is intentionally not covered.

frontend/src/hooks/useJobs.test.ts
```diff
@@ -21,6 +57,42 @@ describe('useJobs', () => {
     });
     await waitFor(() => expect(result.current.isError).toBe(true));
   });
+
+  it('posts to the parent when a refresh-relevant job completes in an iframe', async () => {
```
tylerpotts: Coverage gaps worth closing alongside the change: no post when `window.parent === window`; no post on a `failed` transition (or a post, if you decide failures should trigger a re-scan); the first-observed-completed case; and the always-mounted observer once it exists. Nit: `vi.stubGlobal('parent', ...)` + `vi.unstubAllGlobals()` is the convention elsewhere in the tests and removes the manual `restoreParent` bookkeeping.

tylerpotts (changes requested):
Thanks for the quick turnaround on this half. The message contract matches https://github.com/nebari-dev/jupyterlab-launchpad/pull/87 exactly, the transition-based dedupe is the right shape, and CI is green.

Two changes needed before merge, both small:

1. **The emitter is only alive while the Jobs tab is open.** `useParentJobCompletionMessage` runs inside `useJobs`/`useJob`, and the only production consumer of `useJobs` in the local-mode UI is `components/jobs/Jobs.tsx`, which is rendered inside `<TabsContent value="jobs">` on the workspace detail page. `TabsContent` returns `null` when inactive, so on the default overview tab, and on the Workspaces list page (which mounts no jobs hook at all), nothing is polling jobs and nothing gets posted when the install finishes. Because `previousStatuses` starts empty on mount, a job that completed while the tab was closed is never reported later either: the notification is lost, not delayed. Workspace creation works in your video only because `Workspaces.tsx:181` pre-selects the Jobs tab. Details and a fix sketch inline at `useJobs.ts:48`.

2. **The job-type allowlist excludes jobs that change kernel state.** `env_uninstall` removes `.pixi/envs`, `delete` removes the workspace from `nebi workspace list`, and the package jobs reinstall the env. Inline at `useJobs.ts:6`.

Non-blocking suggestions inline: first-observation handling (becomes important once the observer is always mounted), typing/naming the contract, and test coverage. Once the contract lives in one module, a short embedding note in the docs or `AGENTS.md` would be worth adding; nothing describes it today.

Checked: CI at 46fe19d; fresh clone and `npm ci`; traced every mount of `useJobs`/`useJob` and the `TabsContent` implementation; wrote a throwaway vitest that shows a job first observed as `completed` never posts (its mirror assertion fails with "expected 1 call, got 0"). Line references below are at 46fe19d (the merge of main did not touch the two PR files).

frontend/src/hooks/useHostJobNotifications.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+    server.use(
+      http.get('/api/v1/jobs', () => HttpResponse.json([job({ status })])),
+    );
+
+    const { queryClient } = renderNotifications();
+
+    status = 'completed';
+    await refreshJobs(queryClient);
+
+    expect(postMessage).not.toHaveBeenCalled();
```
tylerpotts, line 106: This assertion can't fail: outside an iframe the query is disabled, so nothing is ever fetched and nothing could reach the spy. With `isEmbedded()` hard-coded to `true` the test still passes, because `status = 'completed'` runs before MSW serves the first request, so the hook's first observation is already `completed` and initialization swallows it. The contract worth pinning here is "no polling when not embedded":

```ts
const handler = vi.fn(() => HttpResponse.json([job({ status: 'running' })]));
server.use(http.get('/api/v1/jobs', handler));

renderNotifications();
await act(() => new Promise((r) => setTimeout(r, 100)));

expect(handler).not.toHaveBeenCalled();
```

Passes at HEAD, fails with `isEmbedded` forced to `true`.

frontend/src/hooks/useHostJobNotifications.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+    await waitFor(() =>
+      expect(queryClient.getQueryData<Job[]>(['jobs'])?.[0].status).toBe(
+        'running',
+      ),
+    );
+
+    status = 'failed';
+    await refreshJobs(queryClient);
+
+    expect(parent.postMessage).not.toHaveBeenCalled();
```
tylerpotts, line 129: Also can't fail as written: TanStack hands the new result to React through a `setTimeout(0)` scheduler, so `await act(invalidateQueries)` resolves before the effect runs, and this assertion executes in that gap (a probe shows the cache already holds `failed` here but the effect hasn't fired yet). With the hook's condition mutated to `job.status !== 'running'` the test stays green. Waiting for the cache first closes the gap:

```ts
status = 'failed';
await refreshJobs(queryClient);
await waitFor(() =>
  expect(queryClient.getQueryData<Job[]>(['jobs'])?.[0].status).toBe('failed'),
);

expect(parent.postMessage).not.toHaveBeenCalled();
```

3/3 runs green at HEAD, 3/3 red under that mutation.

frontend/src/hooks/useHostJobNotifications.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+        jobType: 'env_uninstall',
+        status: 'completed',
+        workspaceId: 'ws-1',
+      },
+      window.location.origin,
+    );
+
+    await refreshJobs(queryClient);
+
+    expect(parent.postMessage).toHaveBeenCalledTimes(1);
```
tylerpotts, line 90: This doesn't pin de-duplication: the second refresh returns data deep-equal to the first, `replaceEqualDeep` hands back the same array, the observer doesn't notify React, and the effect never re-runs. Deleting `previousStatuses.current = nextStatuses;` from the hook leaves the whole file green. Make the second response differ so the effect has to run:

```ts
jobs = [job({ status: 'completed' }), job({ id: 'job-2', status: 'running' })];
await refreshJobs(queryClient);
await waitFor(() =>
  expect(queryClient.getQueryData<Job[]>(['jobs'])).toHaveLength(2),
);

expect(parent.postMessage).toHaveBeenCalledTimes(1);
```

(with the handler serving a `jobs` variable like the last test does). 1 call at HEAD, 2 under the mutation. This is the one I'd most want covered: losing de-dup would make Launchpad refresh on every 2s tick while any job runs.

frontend/src/hooks/useHostJobNotifications.ts
```diff
@@ -0,0 +1,58 @@
...
+import type { JobStatus } from '@/types';
+
+export function useHostJobNotifications(): void {
+  const isLocalMode = useModeStore((s) => s.isLocalMode());
+  const enabled = isEmbedded() && isLocalMode;
+  const previousStatuses = useRef<Record<string, JobStatus>>({});
+  const initialized = useRef(false);
+
+  const { data: jobs } = useQuery({
+    queryKey: ['jobs'],
```
tylerpotts, line 19: Non-blocking: elsewhere conditional consumers thread `enabled` through the resource hook (`useGroups(enabled)`, `useMyGroups(enabled)`, `useCollaborators(id, enabled)`) instead of redefining the query, so the polling policy has one owner. `useJobs({ enabled }: { enabled?: boolean } = {})` plus `const { data: jobs } = useJobs({ enabled })` here would do it, and the two existing callers keep working unchanged. My sketch last round suggested the standalone `useQuery`, so this one's on me.

(Two observers on the same key don't double the request rate, I checked: TanStack re-arms every observer's interval on each query update, so they fire together and the second is deduped.)

frontend/src/hooks/useHostJobNotifications.ts
```diff
@@ -0,0 +1,58 @@
...
+  isEmbedded,
+  NEBI_JOB_COMPLETED_MESSAGE,
+  postToHost,
+} from '@/lib/hostBridge';
+import { useModeStore } from '@/store/modeStore';
+import type { JobStatus } from '@/types';
+
+export function useHostJobNotifications(): void {
+  const isLocalMode = useModeStore((s) => s.isLocalMode());
+  const enabled = isEmbedded() && isLocalMode;
```
tylerpotts, line 14: Worth a comment for the next reader, mirroring `useRemote.ts:50-52`: DSP embeds a local-mode server (`NEBI_MODE: "local"` in https://github.com/nebari-dev/data-science-pack/blob/main/images/nebi/jupyter_server_config.py), team-mode servers are never framed, and the desktop app has `window.parent === window`. Nit while here: `export const useHostJobNotifications = () =>` matches every other module in `hooks/`.

tylerpotts (approved):
Both change requests are addressed and this is the right shape now: a single Layout-mounted observer gated on embedded + local mode, posting for every completed job, with the contract in `lib/hostBridge.ts`. Approving.

Non-blocking, in priority order:

1. **Three of the five tests don't pin the behaviour they're named for.** Details inline at the three assertions, each with a mutation that stays green and a corrected version that goes red. The production code is correct in all three cases (the corrected tests pass at HEAD); this is about regression protection, especially for de-duplication, since losing it would make Launchpad refresh on every 2s tick while any job runs.
2. `useJobs({ enabled })` instead of redefining the `['jobs']` query in the new hook; that's how `useGroups`/`useCollaborators`/the remote hooks do conditional fetching, and it keeps one owner for the polling policy. My sketch last round suggested the standalone `useQuery`, so this one's on me.
3. Comments: one line at the `enabled` gate (DSP runs `nebi serve` with `NEBI_MODE=local`; team-mode servers are never framed) and one at the `Layout.tsx:36` call mirroring the `useRemoteView()` note above it. Nit: `export const useHostJobNotifications = () =>` matches the rest of `hooks/`.
4. Docs: nothing describes the contract. Suggested paragraph for the Frontend section of `AGENTS.md`:

   > **Embedded host bridge.** When Nebi runs in local mode inside a same-origin iframe (JupyterLab Launchpad via jupyter-server-proxy), `useHostJobNotifications()` (mounted once in `Layout.tsx`) posts `window.parent.postMessage({ type: 'nebi:job-completed', jobType, status: 'completed', workspaceId }, window.location.origin)` once per job when its status transitions to `completed`. Host detection and messaging live in `src/lib/hostBridge.ts`; keep Nebi host-agnostic and let the host decide what to do with the message. Cross-origin framing is intentionally unsupported. The payload is an external contract consumed by https://github.com/nebari-dev/jupyterlab-launchpad, so changing it needs coordination there.

5. Please update the PR description; it still describes the `useJobs`/`useJob` version and the job-type allowlist.

The failing "Desktop App / Build (ubuntu-latest)" job is a Go module proxy stream error in "Install wails" (`goldmark-emoji` zip); the same workflow passes on `main`, so it just needs a re-run.

Follow-up worth an issue, not for this PR: `GET /jobs` returns every job with full `logs` and no pagination, and it is now polled every 2s for the life of every embedded session instead of only while the Jobs tab is open. Measured: a non-TTY `pixi install -v` log is a few hundred bytes, so a poll is roughly 0.7 MB per 1,000 jobs, uncompressed, through jupyter-server-proxy and the hub proxy, and nothing prunes jobs. Fine at today's scale; dropping `logs` from the list (the SSE handler already replays stored logs) and adding `?limit=` would cap it for good.

Checked at 26492b1 (the later merge of `main` to a50eba0 doesn't touch the PR files): fresh clone, `npm ci`, 9/9 tests, `tsc`, `biome ci`; six mutations against the hook and bridge (two killed by the transition tests, three survived as described inline, one showing the `isLocalMode` gate is untested); corrected test variants run at baseline and under mutation; traced `Layout` mounting across all routes; confirmed the message constant against the merged Launchpad listener in https://github.com/nebari-dev/jupyterlab-launchpad/pull/87.

tylerpotts:
merging so that I can get this into a new release

PR #546, opened 2026-08-27
Bump Go toolchain to 1.26.7

tylerpotts:
The release workflow's strict toolchain gate (`scripts/check-go-toolchain.sh --strict`) fails on main because go.mod pins go1.26.6 and go1.26.7 is out. This blocks cutting v0.15.

- `go.mod`: toolchain go1.26.6 -> go1.26.7
- `Dockerfile`: golang:1.26.7-alpine with pinned digest

govulncheck on `./cmd/... ./internal/...` is clean.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN npm run build
 # Stage 2: Build Go binary
 # Pinned by exact version + digest; keep in sync with the toolchain directive
 # in go.mod (scripts/check-go-toolchain.sh gates supported lines in CI).
-FROM golang:1.26.6-alpine@sha256:3889b425f035be855a72fb4755265311293b6d414521f0a519d819df32222d83 AS backend-builder
+FROM golang:1.26.7-alpine@sha256:28d89ee9cc0ff9fec75c82ca201e6bf7fdf9a679d4b7b24dfa04f2bb766bb468 AS backend-builder
 WORKDIR /app
 
 # Copy go mod files and download dependencies (cached layer)
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/nebari-dev/nebi
 
 go 1.25.0
 
-toolchain go1.26.6
+toolchain go1.26.7
 
 require (
 	github.com/alicebob/miniredis/v2 v2.37.0


viniciusdc:
expected? 
```
Error: C:\Users\runneradmin\go\pkg\mod\github.com\wailsapp\wails\v2@v2.12.0\cmd\wails\build.go:12:2: github.com/pterm/pterm@v0.12.80: verifying module: github.com/pterm/pterm@v0.12.80: reading https://sum.golang.org/tile/8/0/x125/901: stream error: stream ID 103; INTERNAL_ERROR; received from peer
Error: C:\Users\runneradmin\go\pkg\mod\github.com\wailsapp\wails\v2@v2.12.0\internal\colour\colour.go:7:2: github.com/wzshiming/ctc@v1.2.3: read "https://proxy.golang.org/github.com/wzshiming/ctc/@v/v1.2.3.zip": stream error: stream ID 107; INTERNAL_ERROR; received from peer
Error: Process completed with exit code 1.
```

viniciusdc:
This looks right to me. I checked the two pins independently first:

- `golang:1.26.7-alpine` resolves to exactly `sha256:28d89ee9...bb468`, and the digest it replaces is genuinely `1.26.6-alpine`.
- The new pin is an OCI image index, so it still resolves on both amd64 and arm64 for the multi-arch build.
- Its amd64 config blob reports `GOLANG_VERSION=1.26.7`, same check we did on #510.
- `go.dev/dl?mode=json` currently lists only `go1.27.0` and `go1.26.7` as stable, so `--strict` will pass.

Two things worth folding in, neither blocking.

### 1.26.7 has no security content, and I think the body should say so

No advisory in the Go vuln DB is fixed at 1.26.7. Everything recent on this line (GO-2026-5026, -5942, -6218, -5972) was fixed in 1.26.6. The whole 1.26.7 milestone is one issue, golang/go#80927, "net/http: ReadHeaderTimeout remains active after unencrypted HTTP/2 handoff".

We do set `ReadHeaderTimeout` (`internal/server/server.go:218`, value from `internal/config/config.go:13`), so it looks relevant at a glance, but we aren't exposed. The server uses plain `ListenAndServe`, there's no `TLSConfig` or `NextProtos` anywhere, and `golang.org/x/net` is indirect-only with no http2 import. Stdlib `net/http` only negotiates HTTP/2 over TLS, so that handoff path is unreachable for us.

So this is gate compliance rather than a vulnerability fix, and dropping the `security:` prefix #510 used is the right call. One line in the body saying "bugfix-only release, this is gate compliance" would save the next person the same dig.

### Merge this before #511, and let's cut v0.15 soon

#511 keys its branch on `chore/go-toolchain-<version>` and guards with `git ls-remote --heads origin chore/go-toolchain-1.26.7`. This branch is `bump-go-1.26.7`, so that guard won't match it. If #511 lands first, its next scheduled run sees `go1.26.6` on main and opens a duplicate 1.26.7 PR. Merging this one first makes main current and the bot no-ops.

Also worth remembering `--strict` re-evaluates against go.dev at tag time, not at merge time. Patches have been landing about every ten days lately (1.26.5, 1.26.6, 1.26.7), so if v0.15 slips we'll be doing this again.

<details>
<summary>Smaller notes, nothing to action here</summary>

- The "keep in sync" comment at `Dockerfile:11-12` describes an invariant that's softer than it reads. `GOTOOLCHAIN=local` is baked into the golang image, and under that the `toolchain` directive is a switch instruction, so it degrades to a no-op. Only the `go` line is a hard minimum. I tried it: `go 1.25.0` + `toolchain go1.26.7` built by Go 1.25.0 under `GOTOOLCHAIN=local` exits 0 silently, while setting the `go` line to `1.26.0` correctly errors with `go.mod requires go >= 1.26.0 (running go 1.25.0; GOTOOLCHAIN=local)`. So if the Dockerfile tag ever lags go.mod, the image quietly compiles against the older stdlib. Trivy catches part of that, since its report includes the `app/nebi` gobinary target, but only for fixable HIGH/CRITICAL, so a MEDIUM-only stdlib regression would still ship. Nothing to change here, both pins moved together. I'll raise the reader-side check on #511, which adds a script that rewrites both pin sites but nothing that rejects drift, and whose workflow is schedule plus `workflow_dispatch` so it never runs on a PR anyway.
- #478 doesn't conflict with this. Its Dockerfile hunks are at lines 1-6, 18-27 and 41-47 (node, swag, pixi); line 13 falls between them, so these merge cleanly in either order.
- Unrelated to your change, and I'll open an issue for it: `desktop.yml` also fires on `push: tags: ['v*']` and publishes to the same release via `softprops/action-gh-release` with no `needs:` on `vuln-gate`. A stale-toolchain tag still ships desktop binaries even when goreleaser is blocked, which makes the "Blocks the release" comment at `release.yml:19-21` only half true.
- The real supply-chain gaps in this file are `node:20-alpine` (line 3), `swag@latest` (line 21) and `ghcr.io/prefix-dev/pixi:latest` (line 44), the runtime base that actually ships. #478 pins all three, so nothing to do here.

</details>

viniciusdc (approved):
This looks right to me. I checked the two pins independently first:

- `golang:1.26.7-alpine` resolves to exactly `sha256:28d89ee9...bb468`, and the digest it replaces is genuinely `1.26.6-alpine`.
- The new pin is an OCI image index, so it still resolves on both amd64 and arm64 for the multi-arch build.
- Its amd64 config blob reports `GOLANG_VERSION=1.26.7`, same check we did on #510.
- `go.dev/dl?mode=json` currently lists only `go1.27.0` and `go1.26.7` as stable, so `--strict` will pass.

Two things worth folding in, neither blocking.

### 1.26.7 has no security content, and I think the body should say so

No advisory in the Go vuln DB is fixed at 1.26.7. Everything recent on this line (GO-2026-5026, -5942, -6218, -5972) was fixed in 1.26.6. The whole 1.26.7 milestone is one issue, golang/go#80927, "net/http: ReadHeaderTimeout remains active after unencrypted HTTP/2 handoff".

We do set `ReadHeaderTimeout` (`internal/server/server.go:218`, value from `internal/config/config.go:13`), so it looks relevant at a glance, but we aren't exposed. The server uses plain `ListenAndServe`, there's no `TLSConfig` or `NextProtos` anywhere, and `golang.org/x/net` is indirect-only with no http2 import. Stdlib `net/http` only negotiates HTTP/2 over TLS, so that handoff path is unreachable for us.

So this is gate compliance rather than a vulnerability fix, and dropping the `security:` prefix #510 used is the right call. One line in the body saying "bugfix-only release, this is gate compliance" would save the next person the same dig.

### Merge this before #511, and let's cut v0.15 soon

#511 keys its branch on `chore/go-toolchain-<version>` and guards with `git ls-remote --heads origin chore/go-toolchain-1.26.7`. This branch is `bump-go-1.26.7`, so that guard won't match it. If #511 lands first, its next scheduled run sees `go1.26.6` on main and opens a duplicate 1.26.7 PR. Merging this one first makes main current and the bot no-ops.

Also worth remembering `--strict` re-evaluates against go.dev at tag time, not at merge time. Patches have been landing about every ten days lately (1.26.5, 1.26.6, 1.26.7), so if v0.15 slips we'll be doing this again.

<details>
<summary>Smaller notes, nothing to action here</summary>

- The "keep in sync" comment at `Dockerfile:11-12` describes an invariant that's softer than it reads. `GOTOOLCHAIN=local` is baked into the golang image, and under that the `toolchain` directive is a switch instruction, so it degrades to a no-op. Only the `go` line is a hard minimum. I tried it: `go 1.25.0` + `toolchain go1.26.7` built by Go 1.25.0 under `GOTOOLCHAIN=local` exits 0 silently, while setting the `go` line to `1.26.0` correctly errors with `go.mod requires go >= 1.26.0 (running go 1.25.0; GOTOOLCHAIN=local)`. So if the Dockerfile tag ever lags go.mod, the image quietly compiles against the older stdlib. Trivy catches part of that, since its report includes the `app/nebi` gobinary target, but only for fixable HIGH/CRITICAL, so a MEDIUM-only stdlib regression would still ship. Nothing to change here, both pins moved together. I'll raise the reader-side check on #511, which adds a script that rewrites both pin sites but nothing that rejects drift, and whose workflow is schedule plus `workflow_dispatch` so it never runs on a PR anyway.
- #478 doesn't conflict with this. Its Dockerfile hunks are at lines 1-6, 18-27 and 41-47 (node, swag, pixi); line 13 falls between them, so these merge cleanly in either order.
- Unrelated to your change, and I'll open an issue for it: `desktop.yml` also fires on `push: tags: ['v*']` and publishes to the same release via `softprops/action-gh-release` with no `needs:` on `vuln-gate`. A stale-toolchain tag still ships desktop binaries even when goreleaser is blocked, which makes the "Blocks the release" comment at `release.yml:19-21` only half true.
- The real supply-chain gaps in this file are `node:20-alpine` (line 3), `swag@latest` (line 21) and `ghcr.io/prefix-dev/pixi:latest` (line 44), the runtime base that actually ships. #478 pins all three, so nothing to do here.

</details>

PR #550, opened 2026-08-28
Fix SQLite startup crash dropping workspaces.package_manager on populated databases

### LINKED ISSUE #549: [BUG] - Startup crash on SQLite upgrade: dropping workspaces.package_manager fails with FOREIGN KEY constraint on populated databases
### Summary

Since #522 (`46bf3ab`, 2026-08-26), every SQLite-backed nebi — the desktop app (`app.go`) and the server (`internal/server/server.go`) — crashes on startup when migrating a database created before #522 that contains workspaces referenced by child rows (`jobs`, `publications`, …). The `DropColumn` cleanup for the legacy `workspaces.package_manager` column in `internal/db/db.go` fails with `FOREIGN KEY constraint failed`, aborting startup entirely.

```
[rows:19] INSERT INTO `workspaces__temp`(...) SELECT ... FROM `workspaces`
constraint failed: FOREIGN KEY constraint failed (787)
[rows:0] DROP TABLE `workspaces`
Error: failed to run migrations: failed to drop workspaces.package_manager column: constraint failed: FOREIGN KEY constraint failed (787)
```

### Root cause

SQLite has no native column drop in the driver's emulation path, so [`glebarez/sqlite`](https://github.com/glebarez/sqlite) implements `Migrator().DropColumn` by rebuilding the table: `CREATE TABLE workspaces__temp` → copy rows → `DROP TABLE workspaces` → rename. nebi opens its DSN with `_pragma=foreign_keys(ON)` (`internal/db/db.go`), and with FK enforcement on, `DROP TABLE` on a parent table violates the foreign keys held by `jobs`/`publications` rows.

This is an inconsistency in the driver: `DropTable` and `AlterColumn` wrap their rebuilds in `RunWithoutForeignKey()` (temporary `PRAGMA foreign_keys = OFF`), but `DropColumn` (`migrator.go:151` in v1.7.0) does not.

The existing test `TestMigrateDropsLegacyPackageManagerColumn` passes because it creates a standalone `workspaces` table with no referencing rows — the failure needs a populated DB.

The rebuild runs inside a single transaction, so each failed attempt rolls back cleanly: the database is left intact (no stranded `workspaces__temp` — an earlier version of this issue claimed otherwise, which was wrong), but startup fails identically on every boot with no self-healing.

### Impact

- Desktop app crashes on launch after upgrading past #522 (embedded server never starts; in `wails dev` the UI can silently fall through the Vite proxy to whatever else is on `:8460`).
- Any team server on the default SQLite backend with real usage history fails to boot on upgrade.
- Fresh databases (created after #522) and databases where no job/publication ever referenced a workspace are unaffected, as are Postgres deployments (`DropColumn` is a native `ALTER TABLE ... DROP COLUMN` there).

### Steps to reproduce

1. Run any nebi built before #522 (≤ 2026-08-25) with SQLite; create a workspace and let a job run against it (any `jobs` row referencing the workspace suffices).
2. Upgrade to current `main` and start `nebi serve` (or launch the desktop app) against that same database.
3. Startup aborts with `failed to drop workspaces.package_manager column: constraint failed: FOREIGN KEY constraint failed (787)`.

### Acceptance Criteria

- [ ] The `package_manager` column drop succeeds on a SQLite DB whose `workspaces` rows are referenced by `jobs`/`publications` rows (FK enforcement suspended around the rebuild, pinned to a single connection — the DSN pragma re-enables `foreign_keys` on every new pooled connection, so a bare `PRAGMA foreign_keys = OFF` is not sufficient).
- [ ] `TestMigrateDropsLegacyPackageManagerColumn` (or a sibling test) covers the populated case: workspace + referencing `jobs` row, FK pragma ON.
- [ ] Migration verified end-to-end: a pre-#522 SQLite DB with referencing rows boots on current `main` via both `nebi serve` and the desktop app, with all pre-existing rows intact.

### Out of Scope

- Upstreaming the `RunWithoutForeignKey` fix to `glebarez/sqlite` (worth filing separately, but nebi needs a fix that works with the current pinned driver).
- Postgres migration paths (unaffected).
- A general migration framework / versioned migrations (current GORM auto-migrate approach stays).

jbouder:
Fixes #549

## Problem

Since #522, every SQLite-backed nebi (desktop app and `nebi serve`) crashes on startup when migrating a database created before the `package_manager` removal that contains workspaces referenced by child rows (`jobs`, `publications`, …):

```
Error: failed to run migrations: failed to drop workspaces.package_manager column: constraint failed: FOREIGN KEY constraint failed (787)
```

The glebarez/sqlite driver emulates `DropColumn` by rebuilding the table (create `workspaces__temp` → copy rows → `DROP TABLE workspaces` → rename), and — unlike its `DropTable`/`AlterColumn` paths — does not suspend `foreign_keys` first. With nebi's DSN-level `_pragma=foreign_keys(ON)`, the `DROP TABLE` on the parent violates the foreign keys held by referencing rows and aborts startup. The rebuild runs inside a single transaction, so the failure rolls back cleanly — the database is left intact but cannot boot, and the crash recurs on every startup.

Postgres is unaffected (`DropColumn` is a native `ALTER TABLE ... DROP COLUMN` there).

## Fix

- Run the `package_manager` drop pinned to a single connection (`db.Connection`) with `PRAGMA foreign_keys = OFF`/`ON` around it. Pinning is required because the DSN pragma re-enables enforcement on every new pooled connection, and the pragma must be flipped outside the rebuild's transaction, where SQLite ignores it.
- Keep the plain native drop on Postgres.

An earlier revision of this PR also dropped a "stranded" `workspaces__temp` table at startup, on the theory that failed attempts leave the rebuild's temp table behind. That premise was wrong — the rebuild is transactional and rolls the temp table back on failure (verified empirically and against a real post-crash database) — so that cleanup has been removed.

## Testing

- New `TestMigrateDropsLegacyPackageManagerColumnWithReferencingRows` reproduces the crash scenario: legacy schema with the `NOT NULL` column and a `jobs` row holding an FK to a workspace. It verifies the migration succeeds, the column is gone, and all rows survive.
- Verified end-to-end against a real pre-#522 `nebi.db` (legacy column + referencing job rows): `nebi serve` previously crashed, now boots cleanly with data intact.
- `go test ./internal/db/`, `go build ./internal/...`, `go vet`, `gofmt` all clean.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -100,16 +100,6 @@ func New(cfg config.DatabaseConfig) (*gorm.DB, error) {
 func Migrate(db *gorm.DB, seedRegistry bool) error {
 	slog.Info("Running database migrations...")
 
-	// The SQLite driver rebuilds tables through a workspaces__temp copy; a
-	// failed rebuild (e.g. the package_manager drop below, before it ran
-	// with foreign keys suspended) strands that table with rows already
-	// copied, and any later rebuild's CREATE TABLE would collide with it.
-	if db.Dialector.Name() == "sqlite" {
-		if err := db.Exec("DROP TABLE IF EXISTS `workspaces__temp`").Error; err != nil {
-			return fmt.Errorf("failed to drop stranded workspaces__temp table: %w", err)
-		}
-	}
-
 	// Auto-migrate all models
 	err := db.AutoMigrate(
 		&models.User{},
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -100,6 +100,16 @@ func New(cfg config.DatabaseConfig) (*gorm.DB, error) {
 func Migrate(db *gorm.DB, seedRegistry bool) error {
 	slog.Info("Running database migrations...")
 
+	// The SQLite driver rebuilds tables through a workspaces__temp copy; a
+	// failed rebuild (e.g. the package_manager drop below, before it ran
+	// with foreign keys suspended) strands that table with rows already
+	// copied, and any later rebuild's CREATE TABLE would collide with it.
+	if db.Dialector.Name() == "sqlite" {
+		if err := db.Exec("DROP TABLE IF EXISTS `workspaces__temp`").Error; err != nil {
+			return fmt.Errorf("failed to drop stranded workspaces__temp table: %w", err)
+		}
+	}
+
 	// Auto-migrate all models
 	err := db.AutoMigrate(
 		&models.User{},
@@ -132,7 +142,7 @@ func Migrate(db *gorm.DB, seedRegistry bool) error {
 	// manager, and the column was NOT NULL so leaving it would break inserts
 	// on databases created before its removal.
 	if db.Migrator().HasColumn(&models.Workspace{}, "package_manager") {
-		if err := db.Migrator().DropColumn(&models.Workspace{}, "package_manager"); err != nil {
+		if err := dropLegacyPackageManagerColumn(db); err != nil {
 			return fmt.Errorf("failed to drop workspaces.package_manager column: %w", err)
 		}
 	}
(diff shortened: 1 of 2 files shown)


MUFFANUJ (approved):
LGTM!

MUFFANUJ:
I will wait before merging incase anyone else want to take a look.

Issue #555, opened 2026-08-31
PoC for nebi remote interface


pmeier:
`nebi` currently treats a server different from an OCI registry. Part of the UX rework is to make them peers in terms of the functionality a nebi instance that connects to them. Here is an example interface

```go
import "errors"

type AuthenticationCredentialType string

type AuthenticationCredential struct {
	Type     AuthenticationCredentialType
	Username string                  
	Password string             
	Token    string                     
}

func (ac *AuthenticationCredential) Validate() error {
	switch ac.Type {
	case BasicAuthenticationCredentialType:
		if ac.Username == "" || ac.Password == "" {
			return errors.New("basic auth requires username and password")
		}
	case TokenAuthenticationCredentialType:
		if ac.Token == "" {
			return errors.New("token auth requires token")
		}
	default:
		return errors.New("unknown auth type")
	}
	return nil
}

const (
	BasicAuthenticationCredentialType = "basic"
	TokenAuthenticationCredentialType = "token"
)

type WorkspacePreview struct {
	ID      string
	Name    string
	Version uint
}

type Workspace struct {
	WorkspacePreview
	Data any
}

type Remote interface {
	SupportedAuthentication() []AuthenticationCredentialType
	Authenticate(ac AuthenticationCredential) error
	List() ([]WorkspacePreview, error)
	Pull(id string) (Workspace, error)
	Push(Workspace) error
}
```

1. `SupportedAuthentication() []AuthenticationCredentialType` is used for the client to know which authentication methods to offer to the user. If empty, the client can assume that no authentication is required. That is true for example for public, read-only OCI registries.
2. `Authenticate(ac AuthenticationCredential) error` is used for the initial authentication. If an error is returned, the client should refuse to add the remote.
3. `List() ([]WorkspacePreview, error)` list all workspaces that the authenticated user has access to. Error if the user is not authenticated.
4. `Pull(id string) (Workspace, error)` pulls a workspace from the remote. Error if the user is not authenticated or the workspace doesn't exist or the user has no permissions to read it.
5. `Push(Workspace) error` pushes a new workspace or a new version to the remote. Error if the user is not authenticated, if the user has no write permissions, or if the push would result in a conflicting state (user tries to push version 18, but 18 already exists on the remote). Conflict resolution is out of scope for now.

In the example above `WorkspacePreview` and `Workspace` are just placeholders and need to be filled with the actual fields. We also probably should have some custom error types so the client can better react to the different error reasons.

[...]

PR #556, opened 2026-08-31
Spike: unified Remote interface over nebi server and OCI registry

tylerpotts:
Exploration for https://github.com/nebari-dev/nebi/issues/555: can one `Remote` interface make nebi servers and OCI registries peers from a client's point of view?

**Verdict: yes, with adjustments.** The proposed interface holds up once versions become string refs instead of `uint`, methods take a `context.Context`, and pull gains version addressing. Full findings in `internal/remote/README.md`.

## What this adds

A self-contained spike package, `internal/remote`, alongside the existing code (nothing else is touched):

- `remote.go` — the `Remote` interface, credential types, and typed error sentinels (`ErrNotAuthenticated`, `ErrUnauthorized`, `ErrForbidden`, `ErrNotFound`, `ErrConflict`)
- `oci.go` — `OCIRemote`, wrapping `internal/oci`/oras
- `server.go` — `ServerRemote`, wrapping `internal/cliclient`
- `conformance_test.go` — one shared lifecycle script: discover auth schemes, reject bad credentials, authenticate, push a new workspace, list, pull latest and by ref, push conflict, push v2, typed not-found errors
- `README.md` — the exploration report

## Demonstration

The same conformance script passes unmodified against both backends:

```
go test ./internal/remote/            # OCIRemote vs in-memory Distribution registry (basic-auth + anonymous)
go test -tags e2e ./internal/remote/  # ServerRemote vs real in-process team-mode nebi server (+ token auth)
```

## Key findings (details in the README)

- `internal/oci` has no typed errors; mapping works only because oras errors survive the `%w` chains
- `oci.ListRepositories`/`ListTags` never set `PlainHTTP`, so they can't reach plain-HTTP registries at all (bug worth fixing independently)
- OCI conflict detection is read-before-write (racy); the server's is atomic
- Server workspace creation is async, so push-to-new hides a pending→ready poll behind a synchronous signature
- The server returns 403 for unknown workspace IDs (RBAC before existence), so `ErrNotFound` can't be promised uniformly
- `SupportedAuthentication` is discovery on the server but can only be configuration for OCI; OIDC device flow composes with the interface but can't live inside `AuthenticationCredential`
- Open modeling question: assets (OCI bundles carry arbitrary files; the server push/pull API moves only pixi.toml/pixi.lock)

Deliberately out of scope: wiring into `cmd/nebi` or `/remote/*` handlers, nebi-repo filtering in OCI `List`, Quay token auth, asset transfer.

### DIFF
--- a/internal/remote/server.go
+++ b/internal/remote/server.go
@@ -1,0 +1,242 @@
+package remote
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/nebari-dev/nebi/internal/cliclient"
+)
+
+// ServerRemote implements Remote against a nebi server, wrapping
+// internal/cliclient.
+type ServerRemote struct {
+	// BaseURL is the server root, e.g. "http://localhost:8080".
+	BaseURL string
+	// CreateReadyTimeout bounds how long Push waits for a newly
+	// created workspace to become ready before pushing its version
+	// (workspace creation is asynchronous on the server — a spike
+	// finding). Zero means 60s.
+	CreateReadyTimeout time.Duration
+
+	client *cliclient.Client // set by a successful Authenticate
+}
+
+var _ Remote = (*ServerRemote)(nil)
+
+func (r *ServerRemote) SupportedAuthentication() []AuthenticationCredentialType {
+	// The server additionally supports an OIDC device flow (probed via
+	// GET /auth/device-config); that cannot be expressed as a static
+	// credential and is out of scope for the spike (see README).
+	return []AuthenticationCredentialType{
+		BasicAuthenticationCredentialType,
+		TokenAuthenticationCredentialType,
+	}
+}
+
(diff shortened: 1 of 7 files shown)


internal/remote/server.go
```diff
@@ -0,0 +1,242 @@
...
+	client *cliclient.Client // set by a successful Authenticate
+}
+
+var _ Remote = (*ServerRemote)(nil)
+
+func (r *ServerRemote) SupportedAuthentication() []AuthenticationCredentialType {
+	// The server additionally supports an OIDC device flow (probed via
+	// GET /auth/device-config); that cannot be expressed as a static
+	// credential and is out of scope for the spike (see README).
+	return []AuthenticationCredentialType{
```
pmeier, line 32: @tylerpotts Can you include the device flow in the scope? Maybe we need to split between static and "dynamic" auth.

pmeier:
Closing this as complete. Thanks Tyler!

PR #561, opened 2026-09-03
Split Nebi Into Separate Binaries

### LINKED ISSUE #500: Split into 4 Separate Binaries
### Summary

Split the current monolithic nebi binary into four separate executables:

- `nebi-server`
- `nebi-web`
- `nebi-cli`
- `nebi-desktop`

### Background

The current nebi binary is a hybrid that serves as both a server and a client, controlled by the NEBI_MODE environment variable. This mental model is confusing. Users think they are running a client when in fact they are running a server.

### Definition of Done

1. Four separate binaries build successfully:
  - `nebi-server` - `NEBI_MODE=team nebi serve`
  - `nebi-web` - `NEBI_MODE=local nebi serve`
  - `nebi-cli` - `nebi` without the `serve` subcommand
  - `nebi-desktop` - Wails desktop app
2. `NEBI_MODE` environment variable removed
3. No breaking changes:
   - Database schema unchanged
   - Existing workspaces continue to work
   - Existing CI/CD pipelines still pass
4. Build system updated:
   - Makefile builds all 4 binaries
   - Docker images updated
   - Release process documented
5. Tests pass:
   - Unit tests pass for all binaries
   - E2E tests pass for all binaries
   - No regression in existing functionality
6. Documentation updated:
   - README.md reflects new architecture
   - Command examples updated

### Out of Scope

- Proper client web UI (React app served separately)
- Desktop app features beyond Wails wrapper
- Database migrations
- Breaking changes to API

MUFFANUJ:
## Reference Issues or PRs

Closes #500 

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing features not to work as expected)
- [x] Documentation Update
- [x] Refactoring (no functional changes, no API changes)
- [x] Build related changes
- [x] Other (please describe): Splits Nebi into dedicated CLI, server, local web, and desktop binaries

## Testing

- [x] Did you add new tests?

## Documentation

Updated README, installation docs, CLI docs, server setup docs, component architecture docs, examples, release/conda maintainer docs, and config examples to describe the new split-binary model and replace old `nebi` / `nebi serve` usage with `nebi-cli`, `nebi-server`, `nebi-web`, and `nebi-desktop`.

#### Text styling

#### Non-text content

## Any other comments?

This PR implements the binary split requested in issue #500 by replacing the old mode-switched nebi entry point with dedicated `nebi-cli`, `nebi-server`, `nebi-web`, and `nebi-desktop` executables. It removes `nebi serve` and `NEBI_MODE`, preserves existing CLI functionality under `nebi-cli`, makes server/local runtime mode explicit per binary, and updates Makefile targets, Docker, GoReleaser, CI workflows, installers, config examples, and documentation to match the new structure.


cmd/nebi-cli/publish.go
```diff
@@ -32,10 +32,10 @@ The tag auto-increments (v1, v2, v3, ...) based on existing publications.
 If --registry is not specified, the server's default registry is used.
 
 Examples:
-  nebi publish                                       # publish current directory workspace
-  nebi publish myworkspace
-  nebi publish myworkspace --tag v1.0.0
-  nebi publish myworkspace --repo custom-name --registry ghcr`,
+  nebi-cli publish                                       # publish current directory workspace
```
pmeier: While the folder should be kept with the `nebi-cli` name, we should just leave the actual binary as `nebi` for brevity.

.github/workflows/test-install.yml
```diff
@@ -38,7 +38,10 @@ jobs:
         if: steps.install.outputs.skipped != 'true'
         run: |
           export PATH="$HOME/.local/bin:$PATH"
+          nebi-cli version
           nebi version
+          if command -v nebi-server >/dev/null 2>&1; then nebi-server --help; fi
```
pmeier: We should always test these. In what situations are they not available?

cmd/nebi-cli/e2e_test.go
```diff
@@ -107,9 +107,9 @@ func TestMain(m *testing.M) {
 	serverErr := make(chan error, 1)
 	go func() {
 		serverErr <- server.Run(ctx, server.Config{
-			Port:    port,
-			Mode:    "both",
-			Version: "e2e-test",
+			Port:          port,
+			ComponentMode: "both",
```
pmeier: Why was this renamed?

cmd/nebi-cli/main_test.go
pmeier, line 1: Remove this as this is a one time test and not something we should test on a regular basis.

docs/src/content/docs/installation.md
```diff
@@ -12,19 +12,21 @@ curl -fsSL https://pixi.sh/install.sh | bash
 
 See the [Pixi installation docs](https://pixi.sh) for more options.
 
-## Recommended: Install with Pixi
+## Recommended: Install Released Binaries
```
pmeier, line 15: This needs to be discussed.

.air.toml
```diff
@@ -3,9 +3,9 @@ testdata_dir = "testdata"
 tmp_dir = "tmp"
 
 [build]
-  args_bin = ["serve"]
-  bin = "./tmp/main"
-  cmd = "swag init -g cmd/nebi/serve.go -o internal/swagger --packageName swagger && go build -o ./tmp/main ./cmd/nebi"
+  args_bin = []
+  bin = "./tmp/nebi-server"
+  cmd = "swag init -g main.go -d cmd/nebi-server,internal/api,internal/service,internal/models,internal/limits,internal/metrics,internal/auth -o internal/swagger --packageName swagger && go build -o ./tmp/nebi-server ./cmd/nebi-server"
```
pmeier, line 8: Explain this change please

config.yaml
```diff
@@ -26,5 +27,5 @@ log:
 # pixi_path: /custom/path/to/pixi  # Optional: Custom pixi binary path
 
 storage:
-  environments_dir: ./data/environments  # Directory where environments are stored
```
pmeier, line 29: Was that actually wrong? Workspaces and environments are two different concepts. If it is indeed wrong and should be workspaces, we should also change the directory name everywhere.

Dockerfile
pmeier, line 1: We need a build arg or if that doesn't work a separate Dockerfile for nebi-web as well.

pmeier (changes requested):

.github/workflows/test-install.yml
MUFFANUJ: you are right, removed.

cmd/nebi-cli/e2e_test.go
MUFFANUJ: Renamed it initially to avoid confusion with the newly added `RuntimeMode`: Mode was meant for server/worker / both, while `RuntimeMode` controls team vs local. And again agreed this rename is unnecessary churn, so I will revert it if it looks like it's not a clearer or better choice here

.air.toml
MUFFANUJ: This updates Air to match the split-binary layout. Previously, Air built the combined cmd/nebi binary and ran it with the serve subcommand. After we removed nebi serve, Air now builds cmd/nebi-server and runs that binary directly with no extra args. The Swagger command was updated for the same reason: cmd/nebi/serve.go no longer exists, so Swagger generation now starts from the new server entry point and scans the API packages that contain the annotations. Should i add a comment there about this?

config.yaml
MUFFANUJ: Yes, I checked this against main. The app only reads `storage.workspaces_dir`: the config struct maps `WorkspacesDir` with mapstructure:"workspaces_dir" ([config.go#L124](https://github.com/nebari-dev/nebi/blob/75dfa230616bc798394aefb209ff4c5cdb5c1472/internal/config/config.go#L124)), the default is set as `storage.workspaces_dir` ([config.go#L176](https://github.com/nebari-dev/nebi/blob/75dfa230616bc798394aefb209ff4c5cdb5c1472/internal/config/config.go#L176)), and the env binding is also for `storage.workspaces_dir` / `NEBI_STORAGE_WORKSPACES_DIR` ([config.go#L216-L217](https://github.com/nebari-dev/nebi/blob/75dfa230616bc798394aefb209ff4c5cdb5c1472/internal/config/config.go#L216-L217)). `config.yaml.example` already uses `workspaces_dir` too ([config.yaml.example#L60](https://github.com/nebari-dev/nebi/blob/75dfa230616bc798394aefb209ff4c5cdb5c1472/config.yaml.example#L60). So the root `config.yaml` entry with `environments_dir` ([config.yaml#L29](https://github.com/nebari-dev/nebi/blob/75dfa230616bc798394aefb209ff4c5cdb5c1472/config.yaml#L29)) looks stale and would not configure the storage directory.

.air.toml
pmeier: Please explain the `-d cmd/nebi-server,internal/api,internal/service,internal/models,internal/limits,internal/metrics,internal/auth` to me. Why is this needed now and wasn't before?

.github/workflows/test-install.yml
```diff
@@ -38,10 +38,9 @@ jobs:
         if: steps.install.outputs.skipped != 'true'
         run: |
           export PATH="$HOME/.local/bin:$PATH"
-          nebi-cli version
           nebi version
-          if command -v nebi-server >/dev/null 2>&1; then nebi-server --help; fi
-          if command -v nebi-web >/dev/null 2>&1; then nebi-web --help; fi
+          nebi-server --help
```
pmeier, line 42: Let's also use version for other binaries as test.

Dockerfile
```diff
@@ -33,12 +33,14 @@ RUN make swagger
 # BuildKit to the build's target platform; CI builds each platform on a
 # native runner, so this is always a native compile there.
 ARG VERSION=dev
+ARG NEBI_BINARY=nebi-server
```
pmeier, line 36: No default

Dockerfile
```diff
@@ -33,12 +33,14 @@ RUN make swagger
...
 ARG TARGETOS
 ARG TARGETARCH
-RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
+RUN case "${NEBI_BINARY}" in nebi-server|nebi-web) ;; *) echo "unsupported NEBI_BINARY=${NEBI_BINARY}" >&2; exit 1 ;; esac && \
+    CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
     go build -trimpath \
     -ldflags "-s -w -X main.Version=${VERSION}" \
-    -o /nebi ./cmd/nebi
+    -o /nebi ./cmd/${NEBI_BINARY}
```
pmeier, line 43: This is not a good solution as in the final image, we don't know what binary `nebi` actually presents. If we can dynamically inject the binary with its actual name into the final image, we need to have multiple images. However, we should find a way to not duplicate most of the file. Please explore the options.

docs/src/content/docs/installation.md
pmeier: The default should stay as pixi. You can make a TODO: note into the README that we need to publish the others. However, we don't need to decide this as part of the PR. We are not merging into main and can look at it when everything else is ready.

.air.toml
MUFFANUJ: The explicit internal directory list is not needed here. The old Air command relied on Swag's default scan directory, which is the repo root when Air runs from the repo root. I changed this back to that behavior and only updated the general info file path from the old `cmd/nebi/serve.go` to the new `cmd/nebi-server/main.go` after the binary split.

Dockerfile
MUFFANUJ: Updated this to use explicit Docker targets instead of a `NEBI_BINARY` build arg.

The Dockerfile now shares the frontend/backend setup in common base stages, then has separate final targets for `nebi-server` and `nebi-web`. Each target copies the binary under its real name and uses it as the entrypoint, so the final images are no longer hiding different binaries behind `/app/nebi`.

The workflow now selects `target: nebi-server` or `target: nebi-web` explicitly, and there is no default binary selection anymore.

MUFFANUJ: Should we update the docs to include this? if this seems right i will udpate the docs to include a section for this.

pmeier: Yes please.

MUFFANUJ: Sure.

MUFFANUJ: Done.

pmeier (approved):
Thanks Anuj!

PR #562, opened 2026-09-03
oci: bound manifest, layer, and Quay body reads

### LINKED ISSUE #302: oci: cap io.ReadAll on manifest, layer, and API response bodies
## Context

Follow-up from [#295](https://github.com/nebari-dev/nebi/pull/295) review (@nenb, thanks). Every `io.ReadAll` call in `internal/oci/browser.go` reads whatever the remote end sends without an upper bound. A hostile or misbehaving registry could return a multi-GB body for any of these calls and OOM the client.

Severity is low in practice (Nebi manifests and pixi files are always small, Quay error bodies are short), but capping them is a cheap one-liner per site and removes the whole class of "server exhausts client memory" attack.

## Call sites (4)

Manifest body inside `resolveBundleManifest`:
https://github.com/nebari-dev/nebi/blob/de3447087c6dff8930ac3722845855eaf286e729/internal/oci/browser.go#L355

Manifest body inside `IsNebiRepository` (repository filter peek):
https://github.com/nebari-dev/nebi/blob/de3447087c6dff8930ac3722845855eaf286e729/internal/oci/browser.go#L535

Quay visibility API error body:
https://github.com/nebari-dev/nebi/blob/de3447087c6dff8930ac3722845855eaf286e729/internal/oci/browser.go#L619

Layer body fetch used for pixi.toml and pixi.lock:
https://github.com/nebari-dev/nebi/blob/de3447087c6dff8930ac3722845855eaf286e729/internal/oci/browser.go#L633

## Proposed fix

Replace each site with \`io.ReadAll(io.LimitReader(src, cap))\` using a per-site cap that matches the real expected maximum:

| Site | Purpose | Suggested cap |
|---|---|---|
| L355, L535 | OCI manifest JSON | 1 MiB |
| L619 | Quay API error body | 64 KiB |
| L633 | pixi.toml / pixi.lock body | 16 MiB (pixi.lock can be multi-MB for large solves) |

When the cap trips, return a descriptive error (\`manifest exceeds %d bytes\`) rather than silently truncating.

Could also pull the helper into a shared \`readAllBounded(r, cap, what)\` so the error messages are consistent and the caps are visible in one place.

## Not doing

Changing \`resp.Body.Close\` patterns or the \`io.Copy\` streaming paths in \`ExtractBundle\`, those already stream straight to disk via \`oras.Copy\` and don't buffer.

## Notes

Per [coding convention](https://github.com/nebari-dev/nebi/blob/main/CLAUDE.md) the Quay error body read at L619 should also propagate the status code and body snippet in the returned error, which it already does, just needs the cap added.

MUFFANUJ:
## Reference Issues or PRs

Closes #302 

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [x] Did you add new tests?

#### Text styling

#### Non-text content

## Any other comments?

This PR fixes issue #302 by replacing unbounded network `io.ReadAll` usage in `internal/oci/browser.go` with bounded reads. OCI manifest bodies are now capped at 1 MiB, Quay visibility API error bodies are capped at 64 KiB, and core Pixi layer reads for `pixi.toml` / `pixi.lock` are capped at 16 MiB. The bounded helper reads limit + 1 byte so Nebi can detect oversized responses instead of silently truncating them.
Note: It rejects negative limits up front because the layer path uses `desc.Size` from the remote OCI manifest; valid manifests should never have negative sizes, but malformed or hostile manifests should fail clearly.

The layer read path now rejects declared `pixi.toml` / `pixi.lock` sizes above 16 MiB before fetching, and still rejects responses that send more bytes than the manifest declared. Quay error handling continues to include the HTTP status code and a safe response-body snippet, without allowing a huge error body to be buffered. Regression tests cover oversized manifest bodies, oversized declared core layer sizes, oversized layer transport bodies, and capped Quay error bodies.


internal/oci/browser.go
```diff
@@ -27,6 +28,35 @@ import (
 // tells oras.Copy to skip the fetch (see the PreCopy hook).
 var emptyBlobDigest = digest.FromBytes(nil)
 
+const (
+	maxManifestBytes      int64 = 1 * 1024 * 1024
+	maxQuayErrorBodyBytes int64 = 64 * 1024
+	maxPixiLayerBytes     int64 = 16 * 1024 * 1024
```
tylerpotts: **Blocker.** This duplicates `limits.Defaults().LockBytes` (16 MiB) as a compile-time constant, so raising `limits.lock_bytes` no longer affects import. See the review body for the `PullOptions` direction.

Also: there is no publish-side check. I published a 16,777,217-byte pixi.lock to the test registry and both `PullBundle` and `ExtractBundle` then failed with `pixi.lock layer size 16777217 bytes exceeds cap 16777216 bytes`. Publish path is `publisher.go:262-268`; `assertCoreFile` checks only that the file is regular.

internal/oci/browser.go
```diff
@@ -27,6 +28,35 @@ import (
 // tells oras.Copy to skip the fetch (see the PreCopy hook).
 var emptyBlobDigest = digest.FromBytes(nil)
 
+const (
+	maxManifestBytes      int64 = 1 * 1024 * 1024
```
tylerpotts, line 32: **Blocker.** 1 MiB of manifest is roughly 4,800 asset layers at typical path lengths (about 200 to 245 bytes per descriptor). Nothing in `publisher.go` or `walker.go` limits asset count. I published 6,000 tiny assets: publish succeeded in 1.3 s, the manifest was 1,308,798 bytes, and both pull paths failed with `manifest exceeds 1048576 bytes`. The issue's "manifests are always small" predates asset layers. Either raise this to match what publish permits or add a publish-side guard.

internal/oci/bundle_integration_test.go
```diff
@@ -183,6 +252,128 @@ func TestPullBundle_ListsAssetsWithoutFetchingBytes(t *testing.T) {
...
+	_, err = PullBundle(context.Background(), repoRef, "v1", PullOptions{PlainHTTP: true})
+	if err == nil {
+		t.Fatalf("expected oversized-manifest rejection, got nil")
+	}
+	if !strings.Contains(err.Error(), "manifest exceeds 1048576 bytes") {
+		t.Fatalf("expected manifest cap error, got: %v", err)
+	}
+}
+
+func TestIsNebiRepository_RejectsOversizedManifestBody(t *testing.T) {
```
tylerpotts, line 303: **Blocker.** This test passes with the cap deleted. I replaced the bounded read at `browser.go:601` with plain `io.ReadAll` and it stayed green: the appended `X` bytes make the body invalid JSON, so `IsNebiRepository` returns false regardless. Pad a manifest annotation instead so the body is valid JSON that would classify as Nebi without the cap.

internal/oci/browser.go
```diff
@@ -27,6 +28,35 @@ import (
...
+	maxQuayErrorBodyBytes int64 = 64 * 1024
+	maxPixiLayerBytes     int64 = 16 * 1024 * 1024
+)
+
+type readLimitError struct {
```
tylerpotts, line 37: **Suggestion.** `readLimitError` is unexported and its only `errors.As` consumer flattens it without `%w`. On import, cap errors reach the API as a generic 500 because `handleServiceError` only maps `service.ValidationError`. Same as the pre-existing bundle-size cap, so not a regression, but `pixi.OutputLimitError` and `executor.IsResourceLimitError` show the repo's pattern: export the type or add a predicate and map it in the service.

internal/oci/browser.go
```diff
@@ -27,6 +28,35 @@ import (
...
+	bodyName string
+	maxBytes int64
+}
+
+func (e *readLimitError) Error() string {
+	return fmt.Sprintf("%s exceeds %d bytes", e.bodyName, e.maxBytes)
+}
+
+func readAllBounded(r io.Reader, maxBytes int64, bodyName string) ([]byte, error) {
```
tylerpotts, line 69: **Suggestion.** Returning `data[:maxBytes]` with a non-nil error is an unusual contract and exists only for the Quay snippet. Either document it on the helper or truncate in the Quay caller.

internal/oci/browser.go
```diff
@@ -632,31 +682,39 @@ func ChangeRepositoryVisibility(ctx context.Context, host, repoPath, apiToken st
...
-// fetchLayerBytes returns the full raw bytes for a layer descriptor.
-// The body is bounded to desc.Size+1 so a registry that omits
-// Content-Length and serves an oversized stream cannot exhaust RAM —
-// the size cap in resolveBundleManifest only sees the manifest's
-// declared sizes, not the actual transport bytes.
+// fetchLayerBytes returns the full raw bytes for a core pixi file layer.
+// resolveBundleManifest already rejects declared sizes above maxPixiLayerBytes;
+// this repeats the guard for direct callers, then caps the transport body at
+// desc.Size+1 so oversized responses are caught too.
 func fetchLayerBytes(ctx context.Context, repo *remote.Repository, desc ocispec.Descriptor) ([]byte, error) {
```
tylerpotts: **Nit.** The comment says the repeated guard is "for direct callers", but the only caller is `PullBundle`, which already validated. Say defence-in-depth or drop it.

internal/oci/browser_test.go
```diff
@@ -218,6 +218,48 @@ func TestListRepositoriesViaQuayAPI_FollowsPagination(t *testing.T) {
...
+			t.Errorf("method: got %s want POST", r.Method)
+		}
+		if r.URL.Path != "/api/v1/repository/ns/repo/changevisibility" {
+			t.Errorf("path: got %s", r.URL.Path)
+		}
+		w.WriteHeader(http.StatusBadGateway)
+		_, _ = io.WriteString(w, "quay says no: "+strings.Repeat("X", int(maxQuayErrorBodyBytes)+1024))
+	}))
+	defer srv.Close()
+	http.DefaultClient = srv.Client()
```
tylerpotts: **Suggestion.** Swapping the global `http.DefaultClient` works today because nothing in this package calls `t.Parallel()`, but the package already has the seam for this: `ListRepositoriesViaQuayAPIWithClient` at `browser.go:204`. A `ChangeRepositoryVisibilityWithClient` would match.

internal/oci/bundle_integration_test.go
```diff
@@ -51,6 +53,73 @@ func startTestRegistry(t *testing.T) string {
 	return u.Host
 }
 
+// startManifestRewritingRegistry proxies an inner registry and lets tests
+// replace manifest GET bodies while keeping the rest of the registry behavior.
+func startManifestRewritingRegistry(t *testing.T, innerHost string, rewrite func([]byte) ([]byte, error)) string {
```
tylerpotts: **Nit.** This is a near-copy of the inline blob-rewriting proxy in `TestFetchLayerBytes_RejectsOversizedBody`. Parameterizing on a path predicate would let both tests share one shim.

internal/oci/bundle_integration_test.go
```diff
@@ -183,6 +252,128 @@ func TestPullBundle_ListsAssetsWithoutFetchingBytes(t *testing.T) {
...
+	shimHost := startManifestRewritingRegistry(t, host, func(body []byte) ([]byte, error) {
+		return append(body, []byte(strings.Repeat("X", int(maxManifestBytes)+1))...), nil
+	})
+	repoRef := strings.Replace(res.Repository, host, shimHost, 1)
+
+	_, err = PullBundle(context.Background(), repoRef, "v1", PullOptions{PlainHTTP: true})
+	if err == nil {
+		t.Fatalf("expected oversized-manifest rejection, got nil")
+	}
+	if !strings.Contains(err.Error(), "manifest exceeds 1048576 bytes") {
```
tylerpotts: **Nit.** The payload is built from `maxManifestBytes` but the assertion hardcodes `1048576`. When I mutated the constant to 1 GiB this was the assertion that fired, for the wrong reason. Format the expected string from the constant.

internal/oci/browser.go
```diff
@@ -330,6 +360,20 @@ func classifyBundleManifest(m ocispec.Manifest) (classifiedManifest, error) {
 	return out, nil
 }
 
+func validateCoreLayerSize(desc ocispec.Descriptor) error {
```
tylerpotts: Nice. Rejecting the declared size before the fetch is the right order, and the negative-size guard is cheap insurance against hostile manifests.

tylerpotts (changes requested):
Thanks for taking this on, and for going beyond the one-liner the issue proposed. The shared helper, the fail-before-fetch check on declared sizes, and the negative-size guard are all the right shape, and the pre-existing `exceeds declared size` contract survives untouched.

I'm requesting changes for one root cause that shows up three ways: the new caps are compile-time constants with no publish-side counterpart, so bundles that publish successfully become unpullable, and the operator-tunable `limits.lock_bytes` stops governing the import path.

**What I verified.** Suite green with `-race`, lint clean. I also ran two scenario tests against the in-memory registry and two mutation checks; details in the inline comments.

**The core issue.** `internal/limits` already defines `ManifestBytes` 1 MiB, `LockBytes` 16 MiB, and `MetadataBytes` 64 KiB as configurable defaults (`limits.lock_bytes`, `NEBI_LIMITS_LOCK_BYTES`, documented in `docs/src/content/docs/server-setup.md`). The service enforces `s.limits.LockBytes` on push validation and job admission. This PR hardcodes a second 16 MiB and applies it in `resolveBundleManifest`, which both `PullBundle` and `ExtractBundle` use. An operator who raises `lock_bytes` to 32 MiB now gets imports failing at 16 MiB with no knob. The existing precedent for a cap on this path is `PullOptions.MaxBundleBytes`, supplied by the caller in `internal/service/workspace_import.go:87`.

**Suggested direction.** Add `PullOptions.MaxCoreLayerBytes` (zero means the 16 MiB default so the CLI keeps current behavior), pass `s.limits.LockBytes` from `workspace_import.go`, and either enforce the same cap at publish or add a publish-side manifest-size guard so the round trip is symmetric. Keep the manifest and Quay-error caps as constants; those bound protocol payloads nobody should tune.

One scope note: https://github.com/nebari-dev/nebi/issues/302 explicitly lists the `ExtractBundle` streaming path as "not doing", and this PR now rejects bundles on that path too. I think that's defensible, but please say so in the PR body since it changes `nebi import`.

Out of scope for this PR, but worth a follow-up issue: the buffered pull path never verifies manifest or layer bytes against the resolved digest. oras only checks a digest when `Docker-Content-Digest` is present, and these tests rely on that by stripping the header. `content.ReadAll(r, desc)` from oras-go does size and digest verification and is unused here.

internal/oci/browser.go
```diff
@@ -27,11 +28,66 @@ import (
...
+}
+
+// readAllBounded returns at most maxBytes bytes. When the input exceeds
+// maxBytes it returns that truncated prefix together with readLimitError;
+// callers that need an error snippet can include the returned bytes.
+func readAllBounded(r io.Reader, maxBytes int64, bodyName string) ([]byte, error) {
+	if maxBytes < 0 {
+		return nil, fmt.Errorf("%s has invalid negative limit %d bytes", bodyName, maxBytes)
+	}
+	data, err := io.ReadAll(io.LimitReader(r, maxBytes+1))
```
tylerpotts, line 83: **Blocker.** `maxBytes+1` overflows when `maxBytes` is `math.MaxInt64`, so `io.LimitReader` gets a negative limit, returns EOF immediately, and this function returns an empty slice with a nil error. That is reachable now that the core cap can be disabled: with `limits.lock_bytes: 0` (documented as "disable that specific guard"), `ociCoreLayerLimit` returns -1, `validateCoreLayerSize` accepts any non-negative declared size, and `fetchLayerBytes` passes `desc.Size` straight in here. A registry that declares a pixi.lock layer of size 9223372036854775807 and serves the blob chunked (no Content-Length, so oras does not pre-reject) makes team-mode import succeed and stage an empty pixi.lock. Reproduced in isolation and end to end against the in-memory registry; with the default cap the same manifest is correctly rejected.

Fix: reject `maxBytes >= math.MaxInt64` up front, next to the negative check. Alternatively replace the hand-rolled read in `fetchLayerBytes` with oras `content.ReadAll(reader, desc)`, which enforces exact size and also verifies the digest.

internal/oci/browser.go
```diff
@@ -330,6 +392,51 @@ func classifyBundleManifest(m ocispec.Manifest) (classifiedManifest, error) {
...
+		title = "layer"
+	}
+	if desc.Size < 0 {
+		return fmt.Errorf("%s layer has invalid negative size %d bytes", title, desc.Size)
+	}
+	limit := maxBytes
+	if limit == 0 {
+		limit = defaultMaxCoreBytes
+	}
+	if limit < 0 || desc.Size <= limit {
```
tylerpotts: Part of the Blocker on `readAllBounded`: when `limit < 0` this accepts `desc.Size == math.MaxInt64`, which is the value that overflows downstream. If you keep "negative disables", consider still rejecting sizes at or above `math.MaxInt64 - 1` here so the disabled state means "no operator cap" rather than "any int64".

internal/oci/browser.go
```diff
@@ -382,7 +490,7 @@ func resolveBundleManifest(
 			total += a.Size
```
tylerpotts: **Suggestion.** This sum can overflow negative and pass the `total > opts.MaxBundleBytes` check. Pre-existing and reachable through asset sizes independent of this PR, but the same fix pass should use overflow-safe addition (bail out if `total > math.MaxInt64 - a.Size`).

internal/oci/browser.go
```diff
@@ -59,6 +115,10 @@ type PullOptions struct {
 	// registry cannot exhaust disk by streaming a multi-GB asset blob.
 	// Zero or negative = no cap.
 	MaxBundleBytes int64
+	// MaxCoreLayerBytes caps each buffered core layer (pixi.toml and
```
tylerpotts, line 128: **Suggestion.** `MaxBundleBytes` two fields up treats zero as no cap; this field treats zero as the default and negative as no cap. I asked for the zero-means-default semantic last round so the CLI keeps protection, so I am not reopening it, but exporting the default and having `cmd/nebi/import.go` pass it explicitly would let both fields share one rule and remove the `-1` sentinel from the service.

internal/oci/browser.go
```diff
@@ -27,11 +28,66 @@ import (
 // tells oras.Copy to skip the fetch (see the PreCopy hook).
 var emptyBlobDigest = digest.FromBytes(nil)
 
+const (
+	maxManifestBytes      int64 = 1 * 1024 * 1024
```
tylerpotts, line 35: **Nit.** `maxManifestBytes` bounds the OCI image manifest document; `limits.ManifestBytes` bounds pixi.toml. Same word, different thing. A one-line comment here would stop someone later wiring this to `limits.ManifestBytes` by mistake.

internal/service/workspace_import.go
```diff
@@ -77,9 +77,10 @@ func (s *WorkspaceService) ImportFromRegistry(ctx context.Context, registryID st
 		return nil, &ValidationError{Message: "repository or repository_path is required"}
 	}
 	pullOpts := oci.PullOptions{
-		Username:  ep.Username,
-		Password:  ep.Password,
-		PlainHTTP: ep.PlainHTTP,
+		Username:          ep.Username,
+		Password:          ep.Password,
+		PlainHTTP:         ep.PlainHTTP,
+		MaxCoreLayerBytes: ociCoreLayerLimit(s.limits.LockBytes),
```
tylerpotts: **Suggestion.** Both core layers are capped by `LockBytes` here (one `MaxCoreLayerBytes` for pixi.toml and pixi.lock). Everywhere else the service caps pixi.toml by `ManifestBytes` (1 MiB): `resource_limits.go:30`, `:78`, `workspace_worker.go:57`. So a 5 MiB imported pixi.toml is accepted, `pixi lock` runs against it, and it is only rejected at snapshot time in `CreateVersionSnapshot` with a different message. Still bounded, so not a safety issue, but either pass the tighter cap for pixi.toml (a second option field, or `min(ManifestBytes, LockBytes)`) or add a comment saying the downstream check is intentional.

internal/service/oci_errors.go
```diff
@@ -0,0 +1,17 @@
+package service
+
+import "github.com/nebari-dev/nebi/internal/oci"
+
+func ociCoreLayerLimit(maxBytes int) int64 {
```
tylerpotts: **Nit.** `ociCoreLayerLimit` is a limits conversion living in a file named for errors. Its pixi counterparts `listOptions` and `mapPixiListError` both live in `resource_limits.go`. Either move these two helpers there or rename the file to `oci_limits.go`.

internal/oci/browser.go
```diff
@@ -625,38 +737,42 @@ func ChangeRepositoryVisibility(ctx context.Context, host, repoPath, apiToken st
...
+// the transport body at desc.Size+1 so oversized responses are caught too.
 func fetchLayerBytes(ctx context.Context, repo *remote.Repository, desc ocispec.Descriptor) ([]byte, error) {
 	reader, err := repo.Fetch(ctx, desc)
 	if err != nil {
 		return nil, err
 	}
 	defer reader.Close()
-	limited := io.LimitReader(reader, desc.Size+1)
-	buf, err := io.ReadAll(limited)
+	buf, err := readAllBounded(reader, desc.Size, "layer body")
```
tylerpotts, line 814: Note, non-blocking: because this wraps the `readLimitError` with `%w`, a registry that sends more bytes than its own manifest declared now surfaces to the API client as a 400 ValidationError. That is a registry fault rather than a client one, but the alternative is an opaque 500, so I think this is the better trade. Flagging it so the choice is deliberate.

internal/oci/publisher.go
```diff
@@ -304,6 +315,9 @@ func publishBundle(
 	if err != nil {
 		return PublishResult{}, fmt.Errorf("failed to pack manifest: %w", err)
 	}
+	if err := validateManifestSize(manifestDesc); err != nil {
```
tylerpotts, line 325: Nice. Checking the packed manifest before `remote.NewRepository` means a rejected publish leaves no orphan blobs in the registry, and the test against `registry.invalid` proves it never reaches the network.

internal/oci/bundle_integration_test.go
```diff
@@ -51,6 +53,98 @@ func startTestRegistry(t *testing.T) string {
...
+// startManifestRewritingRegistry proxies an inner registry and lets tests
+// replace manifest GET bodies while keeping the rest of the registry behavior.
+func startManifestRewritingRegistry(t *testing.T, innerHost string, rewrite func([]byte) ([]byte, error)) string {
+	t.Helper()
+	return startBodyRewritingRegistry(t, innerHost, func(r *http.Request) bool {
+		return strings.Contains(r.URL.Path, "/manifests/")
+	}, rewrite)
+}
+
+func oversizedNebiManifest(body []byte) ([]byte, error) {
```
tylerpotts, line 136: This is exactly the right shape. The body stays a valid Nebi manifest, so the only thing standing between the test and a false pass is the bounded read, and swapping it for `io.ReadAll` fails both callers.

internal/service/workspace_publishing.go
```diff
@@ -69,6 +69,7 @@ func (s *WorkspaceService) PublishWorkspace(ctx context.Context, wsID string, re
 		extraTags = append(extraTags, t)
 	}
 
+	maxCoreLayerBytes := ociCoreLayerLimit(s.limits.LockBytes)
```
tylerpotts: Good: computing this once and passing it on both the local and team branches keeps the two modes symmetric, which is the rule this repo asks for around `isLocal` branches.

tylerpotts (changes requested):
Thanks for the thorough follow-up. Every blocker from the first round is fixed, and each fix is pinned by a test that goes red when the code under test is removed. Details of what I ran are at the bottom.

**Verdict.** Requesting changes for one new, narrow issue: an int64 overflow in `readAllBounded` that is reachable once the core-layer cap can be disabled. It is a one-line fix. Everything else below is non-blocking.

**Round-1 dispositions.**
- Hardcoded 16 MiB cap / no publish counterpart: fixed. `PullOptions.MaxCoreLayerBytes` is fed from `limits.LockBytes`, publish enforces the same cap before any network I/O, and the 16,777,217-byte pixi.lock from my original repro is now rejected at publish in under 1 ms with zero registry hits.
- 1 MiB manifest cap with no publish guard: fixed. `validateManifestSize` after `oras.Pack` rejects the 6,000-asset case before the first blob push.
- Vacuous oversized-manifest test: fixed. Padding an annotation keeps the body valid JSON, and replacing the bounded read with `io.ReadAll` now fails both tests.
- Error mapping, injectable client, shared shim, constant-formatted assertions, helper doc comment: all done.
- Not done: the PR body still describes a fixed 16 MiB cap and does not mention that `nebi import` (the `ExtractBundle` path) now rejects oversized bundles, which https://github.com/nebari-dev/nebi/issues/302 listed as out of scope. Please update it. Non-blocking.

**Non-blocking items not tied to a diff line.**
- `internal/api/handlers/registry_browse.go:50` and `:138` build `BrowseOptions` without the new `PlainHTTP` field, so browsing an `http://` registry still fails the TLS handshake. Switching both sites to `ParseRegistryURLFull` and setting the field is four lines and would give the new field a production caller.
- `docs/src/content/docs/server-setup.md:65` describes `lock_bytes` as a stored-lockfile cap. It now also bounds pixi.lock on registry import and on publish; a short clause there would keep the doc complete.
- The publish-time manifest cap is effectively an asset-count ceiling of roughly 4,800 files at typical path lengths. Nothing in the docs mentions it, and `cli-local.md` shows `include = ["src/**", ...]` with no hint that a large tree fails with `manifest size ... exceeds cap`. Worth one sentence in the publish section.
- Still no follow-up issue for the missing digest verification on the buffered pull path (`content.ReadAll` from oras-go does size plus digest and is unused). I will file it unless you want to.

**What I verified at e697c6c.** Suite green with `-race` for `internal/oci` and `internal/service`; golangci-lint and gofmt clean. Six mutations, each observed red at the named assertion: drop the import-side `MaxCoreLayerBytes`, drop the publish-side option, make `validateCoreFileSize` skip the size check, delete `validateManifestSize` at publish, use `io.ReadAll` in `fetchManifestBytes`, make `mapOCILimitError` a passthrough. The overflow below was reproduced both in isolation and end to end against the in-memory registry with a chunked blob response. The three merges from main brought no changes into any PR file, so the two commits since my last review are the complete delta.

internal/service/workspace_import.go
```diff
@@ -80,6 +80,11 @@ func (s *WorkspaceService) ImportFromRegistry(ctx context.Context, registryID st
 		Username:  ep.Username,
 		Password:  ep.Password,
 		PlainHTTP: ep.PlainHTTP,
+		// The OCI reader has one core-layer cap today, so use the
+		// larger lockfile limit while staging imports. pixi.toml is
+		// still checked against ManifestBytes during job admission and
+		// snapshot creation before pixi consumes it.
```
tylerpotts: **Blocker.** This comment is not true at this head, on either half. There is no job-admission check: `ImportFromRegistry` calls `s.Create` with only `ImportStagingDir`, so `validateManifestContent` sees empty content and returns nil, and the create job has no per-job validator (those are wired only for install/update/remove/env jobs). And the snapshot check does not run before pixi consumes the file: the executor seeds the workspace and runs `pixi lock` at `internal/executor/local.go:112-115`, and the first `ManifestBytes` read is `readLimitedTextFile` in `workspace_worker.go:57` during `createVersionSnapshot`, after the solve. So an operator who sets `manifest_bytes: 1048576` gets a full `pixi lock` job spent on a 15 MiB pixi.toml before the workspace fails.

Preferred fix: validate before `s.Create` on both branches. Team mode already has the bytes: `s.validateManifestContent("pixi.toml", result.PixiToml)` and `s.validateLockContent("pixi.lock", result.PixiLock)`. Local mode has them on disk in `stagingDir`: `s.readLimitedTextFile(filepath.Join(stagingDir, "pixi.toml"), "pixi.toml", s.limits.ManifestBytes)` and the lock equivalent. That makes this comment and the doc sentence true and returns a 400 instead of a failed job. Acceptable alternative: rewrite this comment and `server-setup.md:67-68` to say imported pixi.toml is bounded by `lock_bytes` on read and rejected against `manifest_bytes` at snapshot time, after the solve.

docs/src/content/docs/server-setup.md
```diff
@@ -62,7 +62,10 @@ Set a numeric limit to `0` to disable that specific guard. Delete jobs are exemp
 The main job limits are:
 
 - `request_body_bytes`: maximum HTTP request body size.
-- `manifest_bytes`, `lock_bytes`, `metadata_bytes`: maximum stored manifest, lockfile, and metadata sizes.
+- `manifest_bytes`, `lock_bytes`, `metadata_bytes`: maximum stored manifest,
+  lockfile, and metadata sizes. The lockfile cap also bounds OCI registry
+  import and publish of core environment files; imported `pixi.toml` files
+  are still checked against the manifest cap before use.
```
tylerpotts: **Blocker (same item as `workspace_import.go:86`).** "Before use" is defensible only if use means reaching ready status. `pixi lock` has already run on the oversized file by the time the manifest cap is applied. Either add the import-side validation so this becomes literally true, or say the manifest cap is applied at snapshot time after the solve.

docs/src/content/docs/cli-team.md
```diff
@@ -168,6 +168,7 @@ $ nebi publish my-workspace --registry ghcr --tag v1.0
 **Publish** takes the workspace files already on the Nebi server and pushes them to an external OCI registry (e.g., Quay.io, GHCR) for distribution. You must **push** before you **publish**, as publish reads from the server, not your local files.
 
 By default, the content hash tag is used as the primary OCI tag, and a `latest` tag is always created. All workspace tags are propagated to the OCI registry.
+The OCI manifest document is capped at 1 MiB, so extremely large bundle file lists fail before upload.
```
tylerpotts: **Suggestion.** This page describes team-mode `nebi publish`, which goes `PublishWorkspace` (non-local branch) to `oci.PublishWorkspace` to `PublishPixiOnly` to `publishBundle` with a nil asset list. That manifest is config plus two layers and can never approach 1 MiB, so the failure this sentence warns about cannot happen on this path. Either drop it or reword to say team publishes carry only pixi.toml and pixi.lock. The equivalent sentence in `cli-local.md` is accurate.

docs/src/content/docs/cli-local.md
```diff
@@ -127,6 +127,8 @@ exclude = ["*.log", "secrets/**", "notes.md"]
 
 `.gitignore` rules also apply: files git ignores are kept out of
 bundles. Symlinks, device files, and named pipes are skipped silently.
+The OCI manifest document is capped at 1 MiB, so extremely large include
+sets with thousands of small files should be narrowed or packaged separately.
```
tylerpotts, line 131: **Suggestion.** While you are here: the CLI's fixed 16 MiB per-core-layer cap (`oci.DefaultMaxCoreLayerBytes`, used by `nebi import` and local `nebi publish`) is user-visible and documented nowhere. One sentence next to this note would close that gap.

internal/oci/browser.go
```diff
@@ -374,20 +488,58 @@ func resolveBundleManifest(
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func addBundleLayerSize(total, size, maxBytes int64) (int64, error) {
+	if size < 0 {
+		return 0, fmt.Errorf("bundle layer has invalid negative size %d bytes", size)
```
tylerpotts, line 532: **Suggestion.** This negative-size rejection only runs when `MaxBundleBytes > 0`, because `validateBundleSize` returns early otherwise, and the CLI never sets `MaxBundleBytes`. I checked: with the cap unset, a manifest declaring an asset of size -1 passes `PullBundle` and fails later inside `oras.Copy` with a digest mismatch; with the 5 GiB service cap it fails here before any fetch. Fail-safe either way and pre-existing, but manifest validity should not depend on an operator size knob. Moving the `size < 0` check into `classifyBundleManifest` next to the core-layer negative check makes both configurations behave the same and gives the CLI a clear pre-fetch error.

internal/oci/browser.go
```diff
@@ -374,20 +488,58 @@ func resolveBundleManifest(
...
+	}
+	// Check before adding so a hostile manifest cannot wrap the total
+	// negative and slip past the MaxBundleBytes comparison.
+	if total > math.MaxInt64-size {
+		return 0, &sizeLimitError{bodyName: "bundle", size: math.MaxInt64, maxBytes: maxBytes}
```
tylerpotts: **Nit.** Reporting `size: math.MaxInt64` here produces "bundle size 9223372036854775807 bytes exceeds cap 9223372036854775807 bytes" when the cap is MaxInt64, which is exactly the unit test's input. Say the sum overflows int64 rather than fabricating a size.

internal/oci/browser.go
```diff
@@ -27,11 +29,75 @@ import (
...
+// maxBytes it returns that truncated prefix together with readLimitError;
+// callers that need an error snippet can include the returned bytes.
+func readAllBounded(r io.Reader, maxBytes int64, bodyName string) ([]byte, error) {
+	if maxBytes < 0 {
+		return nil, fmt.Errorf("%s has invalid negative limit %d bytes", bodyName, maxBytes)
+	}
+	// The bounded-read pattern below asks LimitReader for maxBytes+1 so
+	// we can detect "one byte too many". Reject the largest int64 value
+	// up front because adding 1 would overflow and make the limit invalid.
+	if maxBytes >= math.MaxInt64 {
```
tylerpotts, line 80: Good. Rejecting the limit before computing `maxBytes+1` closes the exact hole from last round, and the unit test goes red when this line is removed.

internal/oci/browser.go
```diff
@@ -330,6 +402,53 @@ func classifyBundleManifest(m ocispec.Manifest) (classifiedManifest, error) {
...
+	if title == "" {
+		title = "layer"
+	}
+	if desc.Size < 0 {
+		return fmt.Errorf("%s layer has invalid negative size %d bytes", title, desc.Size)
+	}
+	// A disabled operator cap still needs a hard sanity limit: fetchLayerBytes
+	// passes desc.Size to readAllBounded, which reads desc.Size+1 bytes to catch
+	// oversized responses. Values this large would make that addition unsafe.
+	if desc.Size >= math.MaxInt64-1 {
```
tylerpotts, line 419: Good. Putting this before the operator-cap check means a disabled cap still cannot admit a size that would overflow downstream. I verified end to end: with `PullOptions{PlainHTTP: true}` and a manifest declaring a MaxInt64 pixi.lock, this is the check that fires, and removing it together with the `readAllBounded` guard reproduces the empty-lock result.

cmd/nebi/import.go
```diff
@@ -59,8 +59,9 @@ func runImport(cmd *cobra.Command, args []string) error {
 	// policy before any bytes land on disk. This is cheap (one small
 	// GET) and avoids partial-extract state on a rejected destination.
 	peek, err := oci.PullBundle(ctx, repoRef, tag, oci.PullOptions{
-		Concurrency: importConcurrency,
-		PlainHTTP:   plainHTTP,
+		Concurrency:       importConcurrency,
+		PlainHTTP:         plainHTTP,
+		MaxCoreLayerBytes: oci.DefaultMaxCoreLayerBytes,
```
tylerpotts, line 64: Nice. Passing the exported default explicitly makes the CLI's protection visible at the call site instead of relying on a zero-value convention, and both import sites and `PullEnvironment` now do the same.

internal/service/resource_limits.go
```diff
@@ -128,6 +129,13 @@ func (s *WorkspaceService) mapPixiListError(err error) error {
 	return err
 }
 
+func mapOCILimitError(err error) error {
```
tylerpotts, line 132: Good placement, next to `mapPixiListError`. This is now the one file where lower-layer limit errors become `ValidationError`.

tylerpotts (changes requested):

tylerpotts:
Summary for my round-3 review at f638e47 (https://github.com/nebari-dev/nebi/pull/562#pullrequestreview-5270401339). The inline comments carry the specifics; this is the overview.

Thanks for the quick turnaround. The overflow blocker is fixed at both layers and I verified it end to end: a manifest declaring a MaxInt64 pixi.lock with the cap disabled is now rejected with "too-large size", and the same scenario returns an empty lock with a nil error once both new guards are removed, so the guards are exactly what closes the hole. The zero-means-no-cap change is cleaner than what I asked for, and every production pull and publish site now sets a cap explicitly.

**Verdict.** Requesting changes for one item: the comment and doc sentence added in response to my pixi.toml-cap suggestion describe a check that does not happen in that order. Details inline on `workspace_import.go` and `server-setup.md`. Either the small validation fix or a rewording resolves it. Everything else is non-blocking.

**Round-2 dispositions.**
- int64 overflow in `readAllBounded`: fixed, mutation-verified at both guards.
- `MaxBundleBytes` sum overflow: fixed, mutation-verified.
- Tri-state zero semantics: fixed. Zero or negative is no cap everywhere, `oci.DefaultMaxCoreLayerBytes` is exported and passed by the CLI, the service passes `int64(s.limits.LockBytes)` directly.
- `maxManifestBytes` comment, helpers moved out of `oci_errors.go`, `PlainHTTP` wired through both browse handlers: all done.
- pixi.toml capped by `LockBytes`: addressed with a comment whose claim is false. See inline.
- Docs: `server-setup.md` and `cli-local.md` additions are accurate apart from the "before use" clause. The `cli-team.md` sentence is on a page whose publish path never has a file list. See inline.
- PR body: still describes a fixed 16 MiB cap and does not mention that `nebi import` now rejects oversized bundles or that browse honours `http://` registries. Third time asking. Non-blocking, but whoever reads the merge commit will be misled.

**Non-blocking items not on a diff line.**
- `internal/service/workspace_publishing.go:196` is the last `ParseRegistryURL` caller (Quay visibility). Harmless since Quay is HTTPS only, but it will read as a bug once everything else uses `ParseRegistryURLFull`.
- The CLI's fixed 16 MiB core-layer cap is user-visible and appears nowhere under `docs/`. One sentence next to the 1 MiB note in `cli-local.md` would close it.
- Worth stating in `server-setup.md` that `lock_bytes: 0` now disables the OCI core-layer cap on import and publish, since the same knob now governs network reads rather than only stored content.
- Omission semantics still differ by surface: leaving out `WithMaxCoreLayerBytes` yields the 16 MiB default, leaving the struct field unset yields no cap, and the fail-open side is the network one. Every caller is explicit today so I am not reopening it, just recording it.
- Still no follow-up issue for digest verification on the buffered pull path. I will file it.

**What I verified at f638e47.** Suite green with `-race` for `internal/oci`, `internal/service`, and `internal/api/handlers`; golangci-lint and gofmt clean. Mutations observed red at the named assertion: each overflow guard, the bundle-sum guard, the publish default in `resolveConfig`, and reinstating zero-means-default in `validateCoreLayerSize`. The import ordering flagged inline was traced independently three times and I read the executor and worker lines myself.

Issue #582, opened 2026-09-15
research spike: workspace name


pmeier:
The workspace name is currently tightly coupled to the `pixi.toml` manifest. This prevents us from having multiple workspaces with the same name from different sources, e.g. server and OCI registry. The `pixi.toml` does [not require the `name`](https://pixi.prefix.dev/latest/reference/pixi_manifest/#name-optional) to be in there:

> The name of the workspace. If the name is not specified, the name of the directory that contains the workspace is used.

The questions this research spike should answer:

1. Can we detach the workspace name from the `pixi.toml` manifest?
2. If yes, how large is the effort? For this BC of the database is not a concern as long as a script etc. can be run against an old DB to upgrade to the new schema.

pmeier:
Completed by #583 and superseded by #584.

PR #583, opened 2026-09-17
Research spike: decouple Nebi workspace names from Pixi manifests

MUFFANUJ:
This research spike addresses #582 by investigating whether Nebi workspace names can be independent of `pixi.toml` and demonstrating the implementation changes required. The answer is yes: Nebi already stores workspace `UUIDs` separately from non-unique names. Independent display names therefore fit the existing data model. The core implementation does not require a new identity system, database schema migration, or OCI artifact format change.

The prototype makes the stored Nebi name authoritative after registration. It removes automatic manifest-to-database name synchronization from workspace listing, status, pruning, origin lookup, and fallback name resolution. Editing a manifest no longer silently changes the Nebi label. Initial local registration accepts an explicit `--name`; otherwise, it uses the optional manifest name as a one-time suggestion, falling back to the directory name when absent. Manifest parsing remains separate from label selection, so nameless manifests are accepted while malformed TOML is rejected before registration.

CLI pull and OCI import also accept `--name` for newly tracked copies. Their default labels come from the server workspace name and the final OCI repository component, respectively. Already tracked directories retain their existing Nebi labels. Explicit labels are validated before network or filesystem work, and derived labels are checked before files are transferred. These naming operations do not rewrite the downloaded manifest.

Workspace selection now rejects ambiguous names in single-result lookups instead of choosing the first matching record. Explicit selection uses `id::<uuid>`, with `id::<uuid>:<tag>` for tagged remote references. Bare UUID-shaped strings remain display names. The selector is supported across remote lookups and the relevant local commands, while existing interactive duplicate-name pickers remain available. Push creates a workspace only after a genuine name-not-found result; ambiguity, an unknown explicit ID, and server errors do not trigger creation.

A local workspace rename command demonstrates metadata-only renaming. It updates the Nebi label without changing the UUID, directory, manifest, lockfile, or version history. Rename is restricted to ready or failed workspaces with a persisted absolute path and no pending or running jobs. The SQLite transaction checks these conditions under a writer lock. Worker size updates and origin saves now update only their own fields, preventing a stale workspace object from overwriting a newer label.

The UI proof adds a separate Nebi workspace name input to workspace creation. The manifest may contain a different Pixi name or omit it entirely. The shared manifest editor no longer presents a field that displays the Nebi name while writing changes into TOML; optional Pixi names remain editable in TOML mode. Generated pull commands use workspace IDs to distinguish duplicate labels. This spike does not add a server rename endpoint or a UI rename control.

### some visuals

https://github.com/user-attachments/assets/73d4e372-8f98-4ad6-9c03-1a2731fd98bf

The effort is small for the core separation and medium for a complete production rollout. The existing schema already provides the necessary identity and naming fields, and the prototype demonstrates the main CLI and UI changes. Remaining production work concerns consistent validation across entry points, durable source tracking, duplicate-name presentation, and any desired server/UI rename flow. Reopening a database created by the inspected baseline preserved its existing identity and name without a schema change. Additional provenance fields or legacy path backfills would require a separately scoped migration.

Known limitations: Saved origins still resolve through mutable `OriginName` values despite storing `OriginID`. Implicit operations can fail on duplicate names or select another workspace after the original source is renamed. The existing origin-saving code also uses the current directory when pull can write elsewhere through -o. These issues are deliberately deferred and require stable remote identity bound to the correct server and destination.


MUFFANUJ:
This has not been completely reviewed by me yet, since this is just for research and trial purposes. Once the discussion on this is completed, I will go back and do the fixing/reviewing/adding things

pmeier:
Thanks Anuj! Following the discussion in our latest sync, we are going with the full rename as describen in #584.

